PR #49, opened 2025-11-03
add request validity checks + tests

junkicide:
addresses #36


junkicide:
oh I see, can do that over next weekend if not time-sensitive

PR #76, opened 2025-12-11
fix: pass block_height in task metadata for deterministic validation

RonTuretzky:
## Summary
- Adds `block_height` field to `GasKillerTaskData` to enable deterministic storage_updates computation
- Router captures the block height when computing storage_updates and includes it in task metadata
- Validator nodes use this block height to fork at the exact same block, ensuring hash consistency

## Problem
BLS signature verification was failing because the router and nodes computed `storage_updates` at different block heights. When blockchain state changes between computations, different storage updates are produced, resulting in hash mismatches.

## Solution
The router now pins the block height in the task metadata. Validators extract and use this height to ensure they fork at the same block as the router, producing identical `storage_updates` and matching hashes.

## Changes
- `common/src/task_data.rs`: Added `block_height: u64` field with serialization support
- `common/src/validator.rs`: Updated `analyze_transaction` to accept optional block_height param
- `router/src/creator.rs`: Capture and propagate block_height in task metadata

## Test plan
- [x] Unit tests pass (`cargo test`)
- [x] Local build compiles successfully
- [ ] E2E test with rebuilt Docker images (Docker rebuild in progress)


common/src/validator.rs
```diff
@@ -66,22 +68,24 @@ impl GasKillerValidator {
...
         value: Option<alloy::primitives::U256>,
-    ) -> Result<Vec<u8>> {
+    ) -> Result<(Vec<u8>, u64)> {
         let result = Self::analyze_transaction(
             &self.fork_rpc_url,
             contract_address,
             call_data,
             from_address,
             value,
+            None, // Router computes at latest block
```
bagelface: need to pass block height here

PR #77, opened 2025-12-12
Set block height in E2E

bagelface:
This PR is downstream from https://github.com/BreadchainCoop/gas-killer-router/tree/restaking-monorepo so it includes all the commits in that branch. In addition, we added "block height" to task data and point to a new version of "gas analyzer" to ensure deterministically generating storage updates in the validator.

### DIFF
--- a/scripts/trigger_gas_killer.rs
+++ b/scripts/trigger_gas_killer.rs
@@ -41,9 +41,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
         let call_data = hex::decode(call_data_hex.trim_start_matches("0x"))?;
 
         // Get RPC URL for fetching block number if needed
-        let rpc_for_block = env::var("HTTP_RPC")
-            .or_else(|_| env::var("GAS_ANALYZER_RPC"))
-            .map_err(|_| "HTTP_RPC or GAS_ANALYZER_RPC required to fetch block number")?;
+        let rpc_for_block =
+            env::var("HTTP_RPC").map_err(|_| "HTTP_RPC required to fetch block number")?;
         let rpc_url_for_block = Url::parse(&rpc_for_block)?;
         let provider_for_block = ProviderBuilder::new().connect_http(rpc_url_for_block);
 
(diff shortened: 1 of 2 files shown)


common/src/validator.rs
```diff
@@ -62,26 +64,48 @@ impl GasKillerValidator {
         &self.fork_rpc_url
     }
 
+    /// Returns the actual block number to use:
+    /// - If block_number is Some(n) and n > 0, returns n.
+    /// - If block_number is None or Some(0), fetches latest block number from the node.
+    pub async fn resolve_block_number(&self, block_number: Option<u64>) -> Result<u64> {
+        match block_number {
+            Some(n) if n > 0 => Ok(n),
+            _ => {
```
RonTuretzky: I think we may want to fully fail if block number is not being passed in tbh

bagelface:
updated to require "block height" be specified in the requests

scripts/trigger_gas_killer.rs
```diff
@@ -40,26 +40,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
         // Decode hex inputs to bytes
         let call_data = hex::decode(call_data_hex.trim_start_matches("0x"))?;
 
+        // Get RPC URL for fetching block number if needed
+        let rpc_for_block = env::var("HTTP_RPC")
+            .or_else(|_| env::var("GAS_ANALYZER_RPC"))
```
RonTuretzky: where is this `GAS_ANALYZER_RPC` var coming from? we should be re-using the HTTP_RPC

scripts/trigger_gas_killer.rs
```diff
@@ -184,6 +200,18 @@ fn env_var(name: &str) -> Result<String, Box<dyn std::error::Error + Send + Sync
     env::var(name).map_err(|_| format!("{} environment variable is required", name).into())
 }
 
+/// Resolves the block height to use for deterministic execution.
+async fn resolve_block_height<P: Provider>(
```
RonTuretzky: can we remove this func outright?

RonTuretzky (changes requested):
small changes close to merge

PR #97, opened 2026-02-01
test: trigger router multiple times in helm integration test

RonTuretzky:
## Summary
- Update helm integration test to trigger the router 3 times instead of once
- Verify `stateTransitionCount` increments correctly after all triggers
- Check that `SumCalculated` events are emitted for each trigger
- Report success/failure for each individual trigger

## Test plan
- [ ] Helm integration test runs in CI
- [ ] All 3 triggers complete successfully
- [ ] stateTransitionCount and event verification passes


helm/gas-killer/values.yaml
```diff
@@ -126,7 +126,7 @@ router:
   ingressPort: 8080
   ingress:
     enabled: true
-    timeoutMs: 180000
+    timeoutMs: 600000  # 10 minutes
```
bagelface: Bumping `timeoutMs` from 180s to 600s (10 minutes) is a big increase. This is the ingress HTTP timeout — it means a single trigger request can hold an HTTP connection open for 10 minutes.

Is this needed for production, or just to give CI more headroom? If it's CI-only, consider overriding it in the CI helm install step (`--set router.ingress.timeoutMs=600000`) rather than changing the default for all environments.

bagelface (commented):
Nice PR — fixes two real bugs (hardcoded round=0 causing nodes to reject repeat tasks, and build timing causing router timeout) and adds proper multi-trigger verification. The round counter approach is clean and the test is well-structured.

A few minor items:

PR #152, opened 2026-03-30
Push semver images to ghcr.io

bagelface:
- Update ghcr push workflow to use semantic versioning managed by releases
- Update helm values to target pinned images for "node" and "router"


bagelface (commented):
## Review

The semver tagging strategy is solid and the local-dev override pattern in the comments is clean. A few things worth flagging:

### `node-latest` + `pullPolicy: Always` as the default is risky

Setting `tag: node-latest` and `pullPolicy: Always` in `values.yaml` means every pod restart will pull the current latest image. In practice this creates two problems:

1. **Version skew during rollouts** — if a new image is pushed to `node-latest` mid-rollout, some pods start on the old image and some on the new one, with no easy way to tell from the pod spec which is which.
2. **Race between CI and deploy** — a `helm upgrade` triggered right after a merge to main can race against the CI build and pull the previous `latest`.

The SHA tag is already being pushed for every event, which is the safe production anchor. Consider making the default `tag: ""` (or a placeholder like `REPLACE_ME`) and requiring operators to explicitly set a SHA or semver tag, rather than silently defaulting to the floating `latest`. Alternatively, use `pullPolicy: IfNotPresent` with `node-latest`, which at least avoids the mid-rollout skew at the cost of needing manual pod deletions to pick up updates.

### Non-semver release tags will produce garbage MAJOR/MINOR tags

The `sed 's/\.[^.]*$//'` chain works correctly for `v1.2.3` but silently misbehaves for non-semver release names (e.g. `2026-03-31`, `release-1`):

- `2026-03-31` → MINOR: `2026-03`, MAJOR: `2026`
- `release-1` → MINOR: `release`, MAJOR: `` (empty string → `node-` tag)

An empty MAJOR would produce `${IMAGE}:node-` which is a valid but nonsensical tag. A simple guard would prevent this:

```bash
if echo "$VERSION" | grep -qE '^v?[0-9]+\.[0-9]+\.[0-9]+'; then
  # compute MINOR and MAJOR
else
  echo "tags=${SHA_TAG},${IMAGE}:node-${VERSION}" >> $GITHUB_OUTPUT
fi
```

### Removal of `pull_request` trigger

Dropping the `pull_request` trigger means Docker images are no longer built for PRs. This is a reasonable tradeoff (no more `pr-<number>` image clutter), but it means PR authors can't pull and test a built image before merge — they're back to local `kind load` only. If integration testing against a real image becomes important later, a separate PR-scoped workflow or manual `workflow_dispatch` job would fill the gap.

### Minor: branch tags on `dev`/`staging` could conflict with semver

A push to a branch named `v1.2.3` (unlikely but possible) would produce `node-v1.2.3` via the `else` branch and potentially shadow a release tag. Low probability but worth noting if the team uses version-named branches.

RonTuretzky (approved):
lgtm

PR #154, opened 2026-03-31
feat: Helm L2 parity for live Gnosis deployments

RonTuretzky:
## Summary
- Pass `L2_HTTP_RPC` to nodes and router in TESTNET mode so they can validate and execute cross-chain tasks on a live Gnosis chain
- Add `l2_avs_deploy.json` creation to the bridge job (queries `blsApkRegistry` from deployed contracts, no user input needed)
- Add optional YieldDistributor configuration in the bridge job when `yieldDistribution.contractAddress` is set (calls `setAvsAddress`/`setBlsSignatureChecker` using bridge-deployed addresses)
- Update `docker-compose.l2.yml` with full local Anvil fork support: adds `gnosis` service, `yield-distribution` service (voting, time advancement, verification), and wires all services to the local L2

## Test plan
- [ ] Deploy with `global.environment=TESTNET` and `l2.httpRpc` set — verify nodes and router receive `L2_HTTP_RPC`
- [ ] Verify bridge creates `l2_avs_deploy.json` with correct addresses from `l2-deploy.json`
- [ ] Test optional YieldDistributor update by setting `yieldDistribution.contractAddress`
- [ ] Run `docker-compose -f docker-compose.yml -f docker-compose.l2.yml up` and verify full local flow

### DIFF
--- a/docker-compose.l2.yml
+++ b/docker-compose.l2.yml
@@ -1,35 +1,53 @@
-# Override for running gas-killer against a target L2 chain.
+# Override for running gas-killer against a local L2 Anvil fork (Gnosis).
 #
 # Usage:
 #   docker-compose -f docker-compose.yml -f docker-compose.l2.yml up
 #
 # Required additions to .env:
-#   L2_HTTP_RPC=https://rpc.gnosischain.com  # or your preferred L2 RPC
-#   PRIVATE_KEY=0x...                        # must have native gas token on L2 to pay for bridge deployment txns
-#   SET_CONTRACT_VALUES_MANUALLY=true        # set to true to skip YieldDistributor update
+#   L2_FORK_URL=https://rpc.gnosischain.com           # RPC to fork for local L2 Anvil
+#   YIELD_DISTRIBUTOR_ADDRESS=0x37A9CC16412CDA9b4f9B1ABdedd3e10BAD8A06da
+#   YIELD_DISTRIBUTOR_OWNER_ADDRESS=0x5DD2e7db868e33c03daB85d0C623048A7980634E
+#
+# Optional:
+#   BLS_SIG_CHECK_ADDRESS=0x056A41125bB62168B464ddDECd995e8494448913  # on L2 fork
+#   YIELD_CYCLE_LENGTH=1                                               # test-friendly
+#   YIELD_BLOCK_STALE_MEASURE=100000000                                # disable staleness
 #
 # Flow:
-#   eigenlayer  →  bridge  →  node-1/2/3 + router
-#
-# After the bridge completes, manually call setAvsAddress and setBlsSignatureChecker
-# on your YieldDistributor contract using the addresses from config/.nodes/l2-deploy.json.
-# Then trigger distribution via the router ingress at http://localhost:8080.
+#   eigenlayer  →  bridge (gnosis + eigenlayer)  →  yield-distribution  →  node-1/2/3 + router
 
 services:
+  # ─── L2 Anvil (Gnosis fork) ──────────────────────────────────────────────
+  gnosis:
+    platform: linux/amd64
+    image: ghcr.io/breadchaincoop/ethereum:dev
+    env_file:
+      - .env
+    ports:
+      - "8546:8545"
+    entrypoint: ["/bin/sh", "-c"]
+    command:
+      - |
+        anvil --fork-url ${L2_FORK_URL:-https://rpc.gnosischain.com} --host 0.0.0.0 --port 8545 --code-size-limit 65536 \
+          --timeout 120000 \
+          --retries 5 \
+          --compute-units-per-second 300
+
   # ─── Bridge ──────────────────────────────────────────────────────────────────
-  # Deploys RegistryCoordinatorMimic + BLSSignatureChecker to target L2 chain,
-  # bridging EigenLayer operator state from the local Anvil Sepolia fork.
+  # Deploys RegistryCoordinatorMimic + BLSSignatureChecker to the local L2 Anvil,
+  # bridging EigenLayer operator state from the local L1 Sepolia fork.
   bridge:
     platform: linux/amd64
     image: ghcr.io/breadchaincoop/target-contracts/bridge:latest
     depends_on:
       eigenlayer:
         condition: service_completed_successfully
+      gnosis:
+        condition: service_started
     environment:
       - PRIVATE_KEY=${PRIVATE_KEY}
       - L1_HTTP_RPC=http://ethereum:8545
-      - L2_HTTP_RPC=${L2_HTTP_RPC}
-      - YIELD_DISTRIBUTOR_ADDRESS=${YIELD_DISTRIBUTOR_ADDRESS}
+      - L2_HTTP_RPC=http://gnosis:8545
     volumes:
       - ./config/.nodes:/app/.nodes
     entrypoint:
@@ -46,7 +64,7 @@ services:
         # L1_KEY: Anvil's pre-funded test account — nonce 0 on any fresh Sepolia fork,
         # so its CREATE addresses are guaranteed fresh (no CreateCollision).
         L1_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
-        # L2_KEY: your key — must hold the native token on the L2 chain for L2 contract deployments.
+        # L2_KEY: Anvil's pre-funded account works on the local L2 fork too.
         L2_KEY="$$PRIVATE_KEY"
 
         REGISTRY_COORDINATOR_ADDRESS=$$(jq -r '.addresses.registryCoordinator // .addresses.RegistryCoordinator' /app/.nodes/avs_deploy.json)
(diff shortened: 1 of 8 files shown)


docker-compose.l2.yml
```diff
@@ -1,35 +1,53 @@
...
 #
-# After the bridge completes, manually call setAvsAddress and setBlsSignatureChecker
-# on your YieldDistributor contract using the addresses from config/.nodes/l2-deploy.json.
-# Then trigger distribution via the router ingress at http://localhost:8080.
+# Flow:
+#   eigenlayer  →  bridge (gnosis + eigenlayer)  →  yield-distribution  →  node-1/2/3 + router
 
 services:
+  # ─── L2 Anvil (Gnosis fork) ──────────────────────────────────────────────
+  gnosis:
```
bagelface: The service is named `gnosis` but everything else in this file uses generic "l2" terminology. Hardcoding the chain name here means every downstream reference (`http://gnosis:8545` in node/router overrides, `L2_HTTP_RPC=http://gnosis:8545`) is also chain-specific. Recommend renaming to `l2:` (or `l2-anvil:`) so this file stays chain-agnostic — the fork URL already carries the Gnosis specificity.

docker-compose.l2.yml
```diff
@@ -1,35 +1,53 @@
...
+    platform: linux/amd64
+    image: ghcr.io/breadchaincoop/ethereum:dev
+    env_file:
+      - .env
+    ports:
+      - "8546:8545"
```
bagelface: Port `8546` is hardcoded. Consider `"${L2_PORT:-8546}:8545"` to avoid conflicts when running multiple local stacks.

docker-compose.l2.yml
```diff
@@ -1,35 +1,53 @@
...
+    entrypoint: ["/bin/sh", "-c"]
+    command:
+      - |
+        anvil --fork-url ${L2_FORK_URL:-https://rpc.gnosischain.com} --host 0.0.0.0 --port 8545 --code-size-limit 65536 \
```
bagelface: `--code-size-limit 65536` affects whether contracts can deploy and will need tuning for other chains (Arbitrum etc.). Suggest `${ANVIL_CODE_SIZE_LIMIT:-65536}`. The other flags (`--timeout`, `--retries`, `--compute-units-per-second`) are less critical but would be good to expose for the same reason.

docker-compose.l2.yml
```diff
@@ -1,35 +1,53 @@
...
+  # Deploys RegistryCoordinatorMimic + BLSSignatureChecker to the local L2 Anvil,
+  # bridging EigenLayer operator state from the local L1 Sepolia fork.
   bridge:
     platform: linux/amd64
     image: ghcr.io/breadchaincoop/target-contracts/bridge:latest
     depends_on:
       eigenlayer:
         condition: service_completed_successfully
+      gnosis:
+        condition: service_started
```
bagelface: `service_started` only means the container launched, not that Anvil has finished loading the fork. Forking Gnosis mainnet (45M+ blocks) can take several seconds, during which `cast` calls will fail. Either add a healthcheck to the `gnosis` service (e.g. `cast chain-id http://localhost:8545`) and use `service_healthy` here, or have the bridge entrypoint explicitly wait/retry the L2 RPC before proceeding.

docker-compose.l2.yml
```diff
@@ -1,35 +1,53 @@
...
     environment:
       - PRIVATE_KEY=${PRIVATE_KEY}
       - L1_HTTP_RPC=http://ethereum:8545
-      - L2_HTTP_RPC=${L2_HTTP_RPC}
-      - YIELD_DISTRIBUTOR_ADDRESS=${YIELD_DISTRIBUTOR_ADDRESS}
+      - L2_HTTP_RPC=http://gnosis:8545
```
bagelface: Same `gnosis` hardcoding as the service name — if the service is renamed to `l2:` this becomes `http://l2:8545`. (Same applies to line 193, 416, 423, 430, 441.)

docker-compose.l2.yml
```diff
@@ -156,57 +174,268 @@ services:
...
+          # ── Create l2_avs_deploy.json for the router ───────────────────────────
+          BLS_APK_REGISTRY=$$(cast call --rpc-url "$$L2_HTTP_RPC" "$$REG_COORD" "blsApkRegistry()(address)")
+          echo "BLSApkRegistry: $$BLS_APK_REGISTRY"
+
+          cat > /app/.nodes/l2_avs_deploy.json <<DEPLOYEOF
+          {
+            "addresses": {
+              "registryCoordinator": "$$REG_COORD",
+              "blsapkRegistry": "$$BLS_APK_REGISTRY",
+              "blsSigCheck": "$$BLS_SIG_CHECK_ADDRESS",
```
bagelface: `blsSigCheck` here uses `BLS_SIG_CHECK_ADDRESS` — the pre-existing on-fork contract. But the Helm `bridge-job.yaml` path uses `BLS_CHECKER_ADDR` (the bridge-deployed `blsSignatureChecker` from `l2-deploy.json`), which is also what gets configured on the YieldDistributor via `setBlsSignatureChecker`. These two paths write different addresses into `l2_avs_deploy.json` for the same field. The Helm path is likely correct — should this use `$$BLS_CHECKER` (from `l2-deploy.json`) instead?

docker-compose.l2.yml
```diff
@@ -156,57 +174,268 @@ services:
...
+              "counter": "0x0000000000000000000000000000000000000000"
```
bagelface: Vestigial `"counter"` field (zero address) copied from the counter example pattern. If any router/node code reads this field and attempts to use it, zero address will cause a silent failure. Should be removed unless something actually consumes it — in which case it needs a real value.

docker-compose.l2.yml
```diff
@@ -156,57 +174,268 @@ services:
...
+          # ── Step 4: Advance time and mine blocks ───────────────────────────────
+          echo ""
+          echo "=== Step 4: Advancing time ==="
+
+          CURRENT_TIMESTAMP=$$(cast block --rpc-url "$$L2_HTTP_RPC" latest -f timestamp 2>&1 | grep -oE '^[0-9]+' | head -1 || echo "1770000000")
+          CURRENT_BLOCK=$$(cast block-number --rpc-url "$$L2_HTTP_RPC" | tr -d '[:space:]')
+          echo "Current timestamp: $$CURRENT_TIMESTAMP"
+          echo "Current block: $$CURRENT_BLOCK"
+
+          TIME_ADVANCE=$$((8 * 24 * 60 * 60))
```
bagelface: 8 days is hardcoded. Since `YIELD_CYCLE_LENGTH` is already an env var, `YIELD_TIME_ADVANCE` (defaulting to `$((8 * 24 * 60 * 60))`) should be too — useful for faster local test iterations or matching different cycle lengths.

helm/gas-killer/templates/bridge-job.yaml
```diff
@@ -162,6 +162,34 @@ spec:
               chmod 644 /app/.nodes/l2-deploy.json
               echo "l2-deploy.json copied to shared volume" | tee -a "$LOG_FILE"
 
+              # ── Create l2_avs_deploy.json for the router ──────────────────────────
+              echo "=== Creating l2_avs_deploy.json ===" | tee -a "$LOG_FILE"
+              L2_DEPLOY="/app/contracts/artifacts/l2-deploy.json"
+              MIMIC_ADDR=$(jq -r '.registryCoordinatorMimic' "$L2_DEPLOY")
+              BLS_CHECKER_ADDR=$(jq -r '.blsSignatureChecker' "$L2_DEPLOY")
+
+              BLS_APK_REGISTRY=$(cast call --rpc-url "$L2_HTTP_RPC" "$MIMIC_ADDR" "blsApkRegistry()(address)" 2>/dev/null || echo "0x0000000000000000000000000000000000000000")
```
bagelface: The `2>/dev/null || echo "0x000..."` fallback means if the `cast call` fails (L2 RPC not ready, contract missing the method, etc.), `l2_avs_deploy.json` is silently written with a zero `blsapkRegistry` address. Downstream router/node code consuming this file will get a bad address with no indication anything went wrong. Should either fail loudly or log a clear warning before continuing.

helm/gas-killer/templates/bridge-job.yaml
```diff
@@ -162,6 +162,34 @@ spec:
...
+              echo "[BRIDGE] BLSApkRegistry: $BLS_APK_REGISTRY" | tee -a "$LOG_FILE"
+
+              printf '{"addresses":{"registryCoordinator":"%s","blsapkRegistry":"%s","blsSigCheck":"%s","counter":"0x0000000000000000000000000000000000000000"}}' \
```
bagelface: Same vestigial `"counter"` zero-address field as in the docker-compose path. Remove it unless something reads it.

helm/gas-killer/templates/bridge-job.yaml
```diff
@@ -162,6 +162,34 @@ spec:
...
+                "$MIMIC_ADDR" "$BLS_APK_REGISTRY" "$BLS_CHECKER_ADDR" \
+                | jq '.' > /app/.nodes/l2_avs_deploy.json
+              chmod 644 /app/.nodes/l2_avs_deploy.json
+              echo "l2_avs_deploy.json created" | tee -a "$LOG_FILE"
+              cat /app/.nodes/l2_avs_deploy.json | tee -a "$LOG_FILE"
+
+              # ── Optional: Update YieldDistributor with bridge-deployed addresses ──
+              if [ -n "$YIELD_DISTRIBUTOR_ADDRESS" ]; then
+                echo "=== Updating YieldDistributor at $YIELD_DISTRIBUTOR_ADDRESS ===" | tee -a "$LOG_FILE"
+                cast send --rpc-url "$L2_HTTP_RPC" --private-key "$PRIVATE_KEY" \
```
bagelface: In TESTNET mode, `PRIVATE_KEY` is the AVS deployer key — almost certainly not the YieldDistributor owner. These `cast send` calls will be submitted with the wrong sender and revert on the owner check, but the script does not check return codes so it will log `[BRIDGE] YieldDistributor updated` and continue as if it succeeded. Either add a separate `YIELD_DISTRIBUTOR_OWNER_KEY` secret (parallel to `yieldDistribution.ownerAddress` which already exists in values.yaml), or check `cast send` exit codes and fail explicitly.

bagelface (commented):
Overall direction is solid — the `l2_avs_deploy.json` creation and TESTNET `L2_HTTP_RPC` wiring are exactly what is needed to unblock #136. Comments below are individual items to address.

PR #157, opened 2026-04-02
Remove hardcoded threshold value

### LINKED ISSUE #155: Retrieve aggregation threshold from EigenLayer smart contracts
## Summary

Tracks upstream fix in [BreadchainCoop/commonware-restaking#135](https://github.com/BreadchainCoop/commonware-restaking/issues/135).

The aggregation threshold is currently hardcoded in \`router/src/main.rs\`:

```rust
let threshold = 3; // hardcoded for now
```

This needs to be retrieved dynamically from the EigenLayer smart contracts before mainnet deployment, since the threshold will vary by quorum and operator set.

## What needs to change

Once the upstream fix lands in `commonware-restaking`:

1. Remove the hardcoded `threshold = 3` from `router/src/main.rs`
2. Query the threshold from the appropriate EigenLayer contract (likely via the registry coordinator or quorum config)
3. Update any dependent configuration (e.g. `AggregationInput`) to use the dynamic value

## Context

The upstream issue originally noted this was blocked on an EigenLayer middleware v2 refactor — that dependency has since been dropped and we are staying on v1, so this can move forward now.

## References

- Upstream issue: [BreadchainCoop/commonware-restaking#135](https://github.com/BreadchainCoop/commonware-restaking/issues/135)
- Roadmap: [gas-killer/roadmap#1](https://github.com/gas-killer/roadmap/issues/1) — listed as a blocker for mainnet
- Code location: `router/src/main.rs:230` (`let threshold = 3;`)

bagelface:
Dynamically pull threshold from service manager by updating to newest version of commonware-restaking library


router/src/main.rs
```diff
@@ -243,9 +243,7 @@ fn main() {
             contributors_map.insert(verifier, verifier_g1);
         }
 
-        // TODO: Retrieve threshold from EigenLayer smart contracts after middleware refactor
-        // https://github.com/BreadchainCoop/commonware-restaking/issues/135
-        let threshold = 3; // hardcoded for now
+        let threshold = quorum_infos[quorum_number].operator_count;
```
RonTuretzky: is this a dupe pr?

bagelface: the other change happens in the "Counter" example which doesn't affect the implementation here, so it needs to be changed in both places

RonTuretzky: gotcha, well we need to do the 2/3 thing here too

Cargo.toml
```diff
@@ -24,11 +24,11 @@ async-trait = "0.1"
 axum = "0.7"
 bytes = "1.10.1"
 clap = "4.5.37"
-commonware-avs-bindings = { git = "https://github.com/BreadchainCoop/commonware-restaking" }
-commonware-avs-core = { git = "https://github.com/BreadchainCoop/commonware-restaking" }
-commonware-avs-node = { git = "https://github.com/BreadchainCoop/commonware-restaking" }
-commonware-avs-router = { git = "https://github.com/BreadchainCoop/commonware-restaking" }
-commonware-avs-eigenlayer = { git = "https://github.com/BreadchainCoop/commonware-restaking" }
+commonware-avs-bindings = { git = "https://github.com/BreadchainCoop/commonware-restaking", branch = "bagelface/aggregation-threshold-from-operator-count" }
```
bagelface: this will be updated to the default branch once the feature branch is merged into it

router/src/executor.rs
```diff
@@ -188,6 +188,33 @@ impl BlsSignatureVerificationHandler for GasKillerHandler {
             }
         };
 
+        // Preflight check: verify stateTransitionCount matches transition_index before calling
+        // send(). The CachedNonceManager increments its counter during prepare() (before
```
RonTuretzky: Seems like this would be better handled in the cachednoncemanager? Have it increment on a call back from the executor?

RonTuretzky (changes requested):

router/src/executor.rs
bagelface: took a different approach, but updated

RonTuretzky (commented):
Lgtm

RonTuretzky (approved):

PR #233, opened 2026-05-20
fix: require INGRESS_PASSWORD for TESTNET deployments with ingress enabled

### LINKED ISSUE #231: Require and document `INGRESS_PASSWORD` for production deployments
The router supports Bearer token authentication on `/trigger` via the `INGRESS_PASSWORD` env var, but it is commented out in `example.env` with no guidance making it required for production:

```
# INGRESS_PASSWORD= # Set a value to enable Bearer token auth on /trigger
```

In a mainnet deployment, an unprotected `/trigger` endpoint allows anyone to submit tasks to the router.

## What needs to happen

1. Add `INGRESS_PASSWORD` as a required secret in the helm chart (alongside `PRIVATE_KEY` / `FUNDED_KEY`) for TESTNET mode
2. Add a startup-time warning (or hard failure) in the router when `INGRESS=true` and `INGRESS_PASSWORD` is not set
3. Document the expected format and rotation process in the deployment runbook

bagelface:
Closes #231

## What

- **Helm**: `secrets.ingressPassword` is now a hard `required` field when `global.environment=TESTNET` and `router.ingress.enabled=true` — `helm upgrade/install` will fail fast if it is omitted
- **Helm**: `INGRESS_PASSWORD` env var is always injected into the router pod under that same condition (previously only injected when the value was non-empty, which was redundant with the required check)
- **Router**: logs a `warn!` at startup whenever `INGRESS=true` and `INGRESS_PASSWORD` is absent, catching unprotected runs outside of Helm (direct binary invocation, local dev)
- **`example.env`**: uncommented `INGRESS_PASSWORD` and added inline rotation guidance

## Why

An unprotected `/trigger` endpoint allows anyone to submit arbitrary tasks to the router. On testnet this is a free compute drain; on mainnet it is a direct attack surface.

## Notes

- LOCAL deployments are unaffected — the `required` guard only fires on TESTNET
- The `warn!` fires even in LOCAL so operators can see it in logs without a hard failure

### DIFF
--- a/.github/workflows/helm-integration-tests.yml
+++ b/.github/workflows/helm-integration-tests.yml
@@ -18,6 +18,8 @@ jobs:
     name: Test with Helm Charts
     runs-on: ubuntu-latest
     timeout-minutes: 120
+    env:
+      CI_INGRESS_PASSWORD: ci-test-password
 
     steps:
       - name: Checkout repository
@@ -150,6 +152,7 @@ jobs:
             --set secrets.privateKey="$PRIVATE_KEY" \
             --set secrets.fundedKey="$FUNDED_KEY" \
             --set secrets.l2ForkUrl="$L2_FORK_URL" \
+            --set secrets.ingressPassword="$CI_INGRESS_PASSWORD" \
             --set node.image.repository=service \
             --set node.image.tag=node-local \
             --set node.image.pullPolicy=Never \
@@ -399,6 +402,7 @@ jobs:
           export AVS_DEPLOYMENT_PATH="./config/.nodes/avs_deploy.json"
           export GAS_KILLER_TRIGGER_URL=http://localhost:8080/trigger
           export PRIVATE_KEY="0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
+          export INGRESS_PASSWORD="$CI_INGRESS_PASSWORD"
 
           # Get ArraySummation address from deployment file
           ARRAY_SUMMATION_ADDRESS=$(jq -r '.addresses.arraySummation' "$AVS_DEPLOYMENT_PATH")
@@ -505,7 +509,7 @@ jobs:
           PAYLOAD="{\"body\":{\"target_address\":\"$YIELD_DISTRIBUTOR\",\"call_data\":$BYTES_ARRAY,\"transition_index\":$INITIAL_COUNT,\"from_address\":\"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266\",\"value\":\"0\",\"block_height\":$BLOCK}}"
           echo "Payload: $PAYLOAD"
 
-          RESPONSE=$(curl -s -w "\n%{http_code}" -X POST -H "Content-Type: application/json" -d "$PAYLOAD" "$ROUTER_URL/trigger")
+          RESPONSE=$(curl -s -w "\n%{http_code}" -X POST -H "Content-Type: application/json" -H "Authorization: Bearer $CI_INGRESS_PASSWORD" -d "$PAYLOAD" "$ROUTER_URL/trigger")
           HTTP_CODE=$(echo "$RESPONSE" | tail -1)
           BODY=$(echo "$RESPONSE" | head -1)
           echo "Router response ($HTTP_CODE): $BODY"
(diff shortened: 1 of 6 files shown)


BreadrichEngels (changes requested):
Overall this correctly addresses all 3 requirements from #231 — the Helm `required` guard, router warning, CI test updates, and docs all look solid. Two issues to fix before merge:

---

**🐛 Bug — `router/src/factories.rs`**

The `warn!` fires unconditionally whenever `INGRESS_PASSWORD` is absent, even when `INGRESS=false`. If ingress isn't enabled the warning is noise and misleading. #231 specifically asked for it to fire when `INGRESS=true` and no password is set:

```rust
let ingress_enabled = env::var("INGRESS").map(|v| v == "true").unwrap_or(false);
if ingress_enabled && ingress_password.is_none() {
    tracing::warn!(
        "INGRESS_PASSWORD is not set — /trigger endpoint is unauthenticated; set INGRESS_PASSWORD in production"
    );
}
```

---

**⚠️ Minor — `example.env`**

`INGRESS_PASSWORD=` (empty value) is subtly misleading — an operator might set it and think auth is enabled, but the router does `.filter(|p| !p.is_empty())` so an empty value still results in no auth. Better to use a placeholder:

```
INGRESS_PASSWORD=change-me  # Required in production. Requests to /trigger must include: Authorization: Bearer <value>
# To rotate: update INGRESS_PASSWORD in your secrets store, then restart the router pod.
```

Also the second comment line is indented with spaces rather than a leading `#` — some env parsers may choke on it.

---

**📝 Follow-up worth tracking**

The `required` guard in `secret.yaml` only fires for `TESTNET`. When mainnet ships, a deployment with `ingress.enabled=true` won't get the hard fail. Fine for now, but worth a follow-up issue when mainnet prep begins.

bagelface:
> Overall this correctly addresses all 3 requirements from #231 — the Helm `required` guard, router warning, CI test updates, and docs all look solid. Two issues to fix before merge:
> 
> **🐛 Bug — `router/src/factories.rs`**
> 
> The `warn!` fires unconditionally whenever `INGRESS_PASSWORD` is absent, even when `INGRESS=false`. If ingress isn't enabled the warning is noise and misleading. #231 specifically asked for it to fire when `INGRESS=true` and no password is set:
> 
> ```rust
> let ingress_enabled = env::var("INGRESS").map(|v| v == "true").unwrap_or(false);
> if ingress_enabled && ingress_password.is_none() {
>     tracing::warn!(
>         "INGRESS_PASSWORD is not set — /trigger endpoint is unauthenticated; set INGRESS_PASSWORD in production"
>     );
> }
> ```
> 
> **⚠️ Minor — `example.env`**
> 
> `INGRESS_PASSWORD=` (empty value) is subtly misleading — an operator might set it and think auth is enabled, but the router does `.filter(|p| !p.is_empty())` so an empty value still results in no auth. Better to use a placeholder:
> 
> ```
> INGRESS_PASSWORD=change-me  # Required in production. Requests to /trigger must include: Authorization: Bearer <value>
> # To rotate: update INGRESS_PASSWORD in your secrets store, then restart the router pod.
> ```
> 
> Also the second comment line is indented with spaces rather than a leading `#` — some env parsers may choke on it.
> 
> **📝 Follow-up worth tracking**
> 
> The `required` guard in `secret.yaml` only fires for `TESTNET`. When mainnet ships, a deployment with `ingress.enabled=true` won't get the hard fail. Fine for now, but worth a follow-up issue when mainnet prep begins.

The bug in `router/src/factories.rs` is not valid. The warn fires as expected.

BreadrichEngels (changes requested):
The `example.env` fix looks good — keeping it commented resolves the empty-value concern.

The `factories.rs` bug is still unaddressed though. The `warn!` still fires unconditionally whenever `INGRESS_PASSWORD` is absent, regardless of whether ingress is actually enabled. This will spam logs in any LOCAL/non-ingress deployment. One line fix:

```rust
let ingress_enabled = env::var("INGRESS").map(|v| v == "true").unwrap_or(false);
if ingress_enabled && ingress_password.is_none() {
    tracing::warn!(
        "INGRESS_PASSWORD is not set — /trigger endpoint is unauthenticated; set INGRESS_PASSWORD in production"
    );
}
```

Fix that and I'll approve.

BreadrichEngels (approved):
Acknowledged — `create_listening_creator_with_server` is only called when ingress is enabled, so the warn is correctly scoped without an explicit `INGRESS` check. My earlier flag was wrong. LGTM, approved. ✅

PR #234, opened 2026-05-20
Document `AvsServiceManagerWrapper` immutable params in `example.env`

### LINKED ISSUE #228: Explicitly set `AvsServiceManagerWrapper` immutable thresholds before deployment
`DeployAvsServiceManagerWrapper.s.sol` reads `QUORUM_THRESHOLD`, `THRESHOLD_DENOMINATOR`, and `BLOCK_STALE_MEASURE` via `vm.envOr(...)` with defaults of 2, 3, and 300 respectively. None of these are exported in `eigenlayer-bls-local/docker/eigenlayer/main.sh` or documented in either `example.env`, so every deployment silently uses the defaults.

These three values are set as `immutable` in the `AvsServiceManagerWrapper` constructor (`AvsServiceManagerWrapper.sol:32-45`) and **cannot be changed after deployment**.

## What needs to happen

1. Decide whether 2/3 majority and 300-block staleness window are the intended production values (they may well be correct — they just need to be explicit)
2. Add the three vars to both `example.env` files (`eigenlayer-bls-local/example.env` and `service/example.env`)
3. Export them in `main.sh` before the `DeployAvsServiceManagerWrapper.s.sol` forge script runs
4. Document what each value means and why it was chosen

## Note

The helm chart's `yield-distribution-job.yaml` has a separate `BLOCK_STALE_MEASURE_OVERRIDE` mechanism that calls `setBlockStaleMeasure()` at runtime — this is a different contract and unrelated to the immutable constructor param above.

bagelface:
## Summary

`QUORUM_THRESHOLD`, `THRESHOLD_DENOMINATOR`, and `BLOCK_STALE_MEASURE` are immutable constructor parameters in `AvsServiceManagerWrapper` — they cannot be changed after deployment. Previously they were undocumented and silently fell through to forge's `vm.envOr(...)` defaults (2, 3, 300).

- Adds a dedicated `AVS Service Manager Wrapper Configuration` section to `example.env` documenting the intended production values (2/3 quorum, 300-block staleness ≈ 1 hour at ~12 s/block) and explaining the immutability constraint
- `service/example.env` is the env source for the `eigenlayer` container when running the full E2E stack via `service/docker-compose.yml`, so these values must be present here as well as in `eigenlayer-bls-local/example.env` (see BreadchainCoop/eigenlayer-bls-local#35)

Closes #228

### DIFF
--- a/example.env
+++ b/example.env
@@ -36,6 +36,22 @@ FORK_URL=https://ethereum-sepolia-rpc.publicnode.com
 # QUORUM_NUMBER=0
 
 # =============================================================================
+# AVS Service Manager Wrapper Configuration
+# =============================================================================
+# These values are passed as immutable constructor parameters when deploying
+# AvsServiceManagerWrapper and CANNOT be changed after deployment.
+#
+# QUORUM_THRESHOLD / THRESHOLD_DENOMINATOR: fraction of total registered
+# operator stake that must sign a task for it to be accepted. 2/3 means at
+# least two-thirds of stake-weighted operators must co-sign.
+#
+# BLOCK_STALE_MEASURE: maximum age (in blocks) of a valid reference block.
+# At ~12 s/block on Ethereum, 300 blocks ≈ 1 hour.
+QUORUM_THRESHOLD=2
+THRESHOLD_DENOMINATOR=3
+BLOCK_STALE_MEASURE=300
+
+# =============================================================================
 # Environment Mode
 # =============================================================================
 ENVIRONMENT=LOCAL


BreadrichEngels (approved):
Covers the `service/example.env` half of #228 for the docker-compose E2E path.

- Values and documentation are consistent with BreadchainCoop/eigenlayer-bls-local#35 ✅
- Placement in the file is sensible ✅
- Clear explanation of immutability and the rationale for each value ✅

Together with #35, all 4 requirements from #228 are satisfied. LGTM ✅

PR #235, opened 2026-05-21
Add AVS metadata endpoint and parameterize EigenLayer quorum config

### LINKED ISSUE #227: Set production values for quorum economic parameters and AVS metadata
Several fields in `config/config.example.json` (mounted into the eigenlayer container at deploy time) and the helm chart's generated `config.json` contain placeholder or test values that get written on-chain at deployment.

## Outstanding items

| Field | Current value | Problem |
|---|---|---|
| `quorum.minimumStake` | `"1"` | 1 wei — effectively no minimum. Combined with the `1 ether` strategy multiplier in `SetupMiddleware.s.sol:77`, any operator can register with a single token. |
| `metadata.uri` | `"metadataURI"` | Placeholder string written on-chain via `IAllocationManager.updateAVSMetadataURI`. Must be a real hosted URL. |
| `quorum.kickBIPsOfOperatorStake` | `10000` (100%) | Needs deliberate review against churn policy. |
| `quorum.kickBIPsOfTotalStake` | `100` (1%) | Needs deliberate review against churn policy. |
| `quorum.maxOperatorCount` | `32` | Confirm against decentralization targets. |

The strategy multiplier (`1 ether`) is hardcoded in `bls-middleware/contracts/script/SetupMiddleware.s.sol:77` and `IncredibleSquaringDeploymentLib.sol:178` — changing `minimumStake` alone isn't enough without also revisiting the multiplier.

## What needs to happen

1. Decide on a real `minimumStake` value (in terms of LST units, accounting for the strategy multiplier)
2. Create and host the AVS metadata JSON at a stable HTTPS or IPFS URL
3. Review kick BIPs and max operator count against production targets
4. Update `config/config.example.json` and the helm chart `configmap.yaml` with production values
5. Parameterize the strategy multiplier in bls-middleware so it can be set via config rather than hardcoded

bagelface:
## Summary

- Adds `GET /avs-metadata` endpoint to the router ingress server, serving the EigenLayer-spec JSON that `updateAVSMetadataURI` points to (name, website, description, logo, twitter, operatorSets)
- Pulls all quorum parameters (`minimumStake`, `maxOperatorCount`, `kickBIPsOfOperatorStake`, `kickBIPsOfTotalStake`) and `metadata.uri` out of hardcoded configmap values and into `values.yaml` under `eigenlayer.quorum`
- `metadata.uri` is now automatically derived from `ingress.host` — no manual URL wiring required
- AVS metadata fields and operator set details are configurable via `router.avsMetadata` in `values.yaml` and threaded through as env vars; `operatorSets` field is omitted when `operatorSet.name` is empty
- Adds production quorum stub with TODOs to `mainnet-overrides.yaml` (closes part of #227)

## Test plan

- [ ] `cargo test -p gas-killer-router` passes (56 tests including new `/avs-metadata` endpoint tests)
- [ ] `helm template` with `ingress.host` set renders correct `metadata.uri` and all `AVS_OPSET_*` env vars
- [ ] `helm template` without `ingress.host` falls back to `metadataURI` placeholder
- [ ] `GET /avs-metadata` returns valid JSON matching the EigenLayer metadata spec
- [ ] Endpoint is accessible without auth even when `INGRESS_PASSWORD` is set

### DIFF
--- a/.github/workflows/helm-integration-tests.yml
+++ b/.github/workflows/helm-integration-tests.yml
@@ -190,7 +190,7 @@ jobs:
           fi
 
           echo "Found setup job: $SETUP_JOB"
-          kubectl wait --for=condition=complete "$SETUP_JOB" --timeout=3600s
+          kubectl wait --for=condition=complete "$SETUP_JOB" --timeout=5400s
 
           echo "Setup job completed successfully"
           kubectl logs "$SETUP_JOB" --tail=30
--- a/.github/workflows/helm-integration-tests.yml
+++ b/.github/workflows/helm-integration-tests.yml
@@ -204,7 +204,7 @@ jobs:
 
           if [ -n "$BRIDGE_JOB" ]; then
             echo "Found bridge job: $BRIDGE_JOB"
-            kubectl wait --for=condition=complete "$BRIDGE_JOB" --timeout=1800s || {
+            kubectl wait --for=condition=complete "$BRIDGE_JOB" --timeout=5400s || {
               echo "Bridge job failed or timed out"
               kubectl logs "$BRIDGE_JOB" --tail=100 || true
               exit 1
--- a/config/config.example.json
+++ b/config/config.example.json
@@ -1,7 +1,7 @@
 {
     "quorum": {
         "minimumStake": "1",
-        "maxOperatorCount": 32,
+        "maxOperatorCount": 4,
         "kickBIPsOfOperatorStake": 10000,
         "kickBIPsOfTotalStake": 100
     },
--- a/example.env
+++ b/example.env
@@ -122,7 +122,16 @@ INGRESS_TIMEOUT_MS=0
 # AVS_METADATA_WEBSITE=https://gaskiller.xyz
 # AVS_METADATA_DESCRIPTION=Verifiable off-chain compute service for EVM smart contracts via EigenLayer
 # AVS_METADATA_LOGO=https://...
-# AVS_METADATA_TWITTER=https://x.com/... 
+# AVS_METADATA_TWITTER=https://x.com/...
+#
+# Operator set entry (omit AVS_OPSET_NAME to exclude the operatorSets field entirely)
+# AVS_OPSET_NAME=Gas Analysis Quorum
+# AVS_OPSET_ID=0
+# AVS_OPSET_DESCRIPTION=BLS-aggregated operator set for verifiable gas analysis tasks
+# AVS_OPSET_SOFTWARE_NAME=gas-killer-node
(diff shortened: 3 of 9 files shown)


BreadrichEngels (changes requested):
Good coverage of the in-scope items from #227. Helm parameterization is clean, the auto-derived `metadata.uri` from `ingress.host` is a nice touch, and the test coverage for the new endpoint (including the no-auth case) is solid.

One real issue to fix, plus a couple of nits:

---

**🐛 `logo` and `twitter` serialize as empty strings when unset — `router/src/ingress.rs`**

`logo` and `twitter` are plain `String` fields. When left unset they'll serialize as `""` in the JSON response:

```json
{ "logo": "", "twitter": "", ... }
```

The EigenLayer indexer likely expects these fields to be absent rather than present as empty strings (inconsistent with how `operatorSets` is handled via `skip_serializing_if = "Option::is_none"`). Should be `Option<String>` with the same treatment:

```rust
#[serde(skip_serializing_if = "Option::is_none")]
pub logo: Option<String>,
#[serde(skip_serializing_if = "Option::is_none")]
pub twitter: Option<String>,
```

And in `factories.rs` read them as `Option`:

```rust
logo: env::var("AVS_METADATA_LOGO").ok().filter(|s| !s.is_empty()),
twitter: env::var("AVS_METADATA_TWITTER").ok().filter(|s| !s.is_empty()),
```

---

**⚠️ `AVS_OPSET_*` vars undocumented in `example.env`**

`AVS_METADATA_*` vars are documented but the `AVS_OPSET_*` vars injected by the Helm chart have no corresponding entries in `example.env`. Anyone running the binary directly (not via Helm) has no reference for them. Worth adding a commented-out block alongside the other AVS metadata vars.

---

**⚠️ Trailing whitespace in `example.env`**

```
# AVS_METADATA_TWITTER=https://x.com/... 
```

Trailing space after `...` — minor but worth cleaning up.

---

**📝 Note on `mainnet-overrides.yaml` comment**

> "These are written on-chain by the setup job and cannot be changed without re-running it."

This is true for the immutable constructor params from #228 (QUORUM_THRESHOLD etc.) but `maxOperatorCount`, kick BIPs, and `minimumStake` in the quorum config may be updatable via governance/admin calls post-deployment depending on the middleware implementation. Worth a more precise comment if that distinction matters.

BreadrichEngels (approved):
All feedback addressed — `logo`/`twitter` are properly `Option<String>` with `skip_serializing_if`, `AVS_OPSET_*` vars are documented in `example.env`, trailing whitespace cleaned up, and the overly broad comment in `mainnet-overrides.yaml` is gone. LGTM ✅

PR #272, opened 2026-06-11
feat: add end-to-end round-latency histogram and dashboard panel

### LINKED ISSUE #265: Add end-to-end round-latency histogram and surface orchestrator metrics on the existing Grafana dashboard
## Problem
The router already has a deployed Grafana round-lifecycle dashboard — `helm/gas-killer/templates/monitoring-grafana-dashboard.yaml` (provisioned via the kube-prometheus-stack sidecar) — with "Latency Breakdown" and "Executor Phase Breakdown" sections covering Aggregation Rounds/min, Execution Duration (p50/p95), Storage Computation, P2P Round-Trip (p50/p99), Node EVMSketch, Task Queue Depth, and the executor phases (chain detection, hash preflight, supportsInterface, TX send, receipt confirmation). The per-phase histograms are defined in `router/src/metrics.rs`.

Two gaps remain for aggregation-speed work:
1. **No single end-to-end round-latency metric.** Round latency must be hand-assembled from `gas_killer_p2p_round_trip_seconds` (creator-dispatch → threshold sigs) + `gas_killer_execution_duration_seconds` (full `handle_verification`, incl. receipt wait). There is no `gas_killer_round_latency_seconds` giving p50/p95/p99 in one query.
2. **Orchestrator round-internal timing is invisible.** Time-to-quorum, signature-arrival spread, and abandoned/timed-out rounds — the key signals for tuning `AGGREGATION_FREQUENCY` — are not exposed; they will be added by BreadchainCoop/commonware-restaking#41.

## Proposed solution
- Add a single end-to-end histogram `gas_killer_round_latency_seconds` (≈ `p2p_round_trip` + `execution_duration` — a convenience aggregate, not a newly-observable quantity). **Caveat:** if it reuses the existing single-slot `DispatchTime` (`router/src/creator.rs:22-27`), it inherits the documented staleness behaviour (a failed round without `handle_verification` bleeds a stale timestamp into the next round); acknowledge this or address it alongside the per-round DispatchTime tracking work. Note the start point is **creator-dispatch**, which is *after* the multi-second storage computation — so this is "dispatch → on-chain confirmation," not "ingress-accept → confirmation."
- **Extend the existing dashboard** (`monitoring-grafana-dashboard.yaml`) with panels for restaking#41's `avs_orchestrator_*` metrics (time-to-quorum, abandoned-round rate, signatures-per-round) once they land — do not create a new dashboard.
- Surface restaking#41's orchestrator metrics on the router `/metrics`. The orchestrator runs inside the commonware runtime, so its metrics register into the runtime Context registry the router already encodes (`router/src/main.rs:59-61` concatenates `context.encode()` + the gas-killer collector) — they should surface there automatically once registered, not via `router/src/metrics.rs`.

## Code locations
- `router/src/metrics.rs` — add `gas_killer_round_latency_seconds`
- `router/src/executor.rs:383-414` — round-completion hook (observe end-to-end here); receipt confirmation itself is at `:326-331`
- `router/src/creator.rs:316-319` — dispatch timestamp (measurement start); single-slot caveat at `:22-27`
- `helm/gas-killer/templates/monitoring-grafana-dashboard.yaml` — add aggregation/orchestrator panels
- `router/src/main.rs:59-61` — `/metrics` handler (orchestrator metrics surface via `context.encode()`)

## Impact / Effort / Risk
- **Impact:** Medium — the dashboard already covers most of the lifecycle; marginal work is the convenience histogram + surfacing #41's orchestrator metrics (the latter gated on #41).
- **Effort:** Low-Medium (1-2 days), partly gated on restaking#41.
- **Risk:** Low (observability only).

## Acceptance Criteria
- `gas_killer_round_latency_seconds` exposed (DispatchTime caveat addressed or noted).
- Existing dashboard extended with orchestrator time-to-quorum / abandoned-round panels once restaking#41 lands.
- Orchestrator metrics confirmed visible on the router `/metrics`.

Depends on BreadchainCoop/commonware-restaking#41 (orchestrator metrics).
Part of gas-killer/roadmap#10

bagelface:
Adds the `gas_killer_round_latency_seconds` end-to-end histogram from #265 and surfaces orchestrator internals on the dashboard.

## Changes
- **`router/src/metrics.rs`** — new `gas_killer_round_latency_seconds` histogram measuring creator dispatch → `verifyAndUpdate` receipt confirmation. Observed for **successful rounds only**: failed rounds have no on-chain confirmation, and a receipt-timeout failure would otherwise inject artificial ~120s samples into the percentiles this metric exists to track.
- **`router/src/executor.rs`** — `handle_verification` stamps dispatch time before the P2P round-trip and observes round latency on the success path.
- **`router/src/factories.rs` / `builder.rs` / `main.rs`** — threads `&context` into `create_gas_killer_executor` and calls `.with_metrics(context)` on the BLS executor so EigenLayer state-retrieval metrics (`executor_operator_cache_misses_total`, etc.) register under the runtime context and surface via `context.encode()` on `/metrics`.
- **`Cargo.toml`** — pins `prometheus-client` to 0.22.3 (0.23 changed the `Histogram::new` API; `.into_iter()` adapters added throughout).
- **`helm/.../monitoring-grafana-dashboard.yaml`** — new **Orchestrator Internals** section (y=32) with six panels:
  - Time to Quorum (p50/p95) — round broadcast → threshold reached, isolates aggregation from EigenLayer reads
  - Signature Arrival Spread (p50/p99) — detects straggler operators holding rounds open
  - Round Outcomes / min — executed / execution failed / window timeout
  - Signatures / min by Outcome — success / invalid / dropped
  - EigenLayer State Retrieval (p50/p95) — operator resolution + non-signer reads between quorum and `handle_verification`
  - Operator Cache Misses — stat panel for cache miss counter
  - Existing Ingress and Executor Phase Breakdown sections shifted down to y=49+.

## Notes on issue #265 scope
- The **DispatchTime single-slot staleness caveat** called out in the issue is already resolved on `main`: dispatch timestamps are keyed per-round with eviction, so the new histogram reuses the same per-round instant with no staleness bleed.
- Orchestrator internals (time-to-quorum, signature-arrival spread, round outcomes) now surface on `/metrics` via `context.encode()` after the `with_metrics` wiring; the new dashboard panels make them immediately visible.
- The start point is creator dispatch (after multi-second storage computation) — this is "dispatch → on-chain confirmation," not "ingress-accept → confirmation." The metric help text and dashboard header state this.

## Testing
- New unit test asserting the histogram is registered and encodes observations.
- `cargo fmt`, `clippy --all-targets --all-features -D warnings`, `check`, `test` all pass.

Closes #265 / part of gas-killer/roadmap#10.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2175,7 +2175,7 @@ checksum = "1d07550c9036bf2ae0c684c4297d503f838287c83c53686d05370d0e139ae570"
 [[package]]
 name = "commonware-avs-bindings"
 version = "0.1.0"
-source = "git+https://github.com/BreadchainCoop/commonware-restaking#07e5a4e9ebe26967b0f57bd83c0328058527e9a2"
+source = "git+https://github.com/BreadchainCoop/commonware-restaking#b20d0a512ae4e95c8e657b4b00403327d83fb86a"
 dependencies = [
  "alloy",
  "alloy-primitives",
@@ -2190,7 +2190,7 @@ dependencies = [
 [[package]]
 name = "commonware-avs-core"
 version = "0.1.0"
-source = "git+https://github.com/BreadchainCoop/commonware-restaking#07e5a4e9ebe26967b0f57bd83c0328058527e9a2"
+source = "git+https://github.com/BreadchainCoop/commonware-restaking#b20d0a512ae4e95c8e657b4b00403327d83fb86a"
 dependencies = [
  "alloy",
  "anyhow",
@@ -2212,7 +2212,7 @@ dependencies = [
 [[package]]
 name = "commonware-avs-eigenlayer"
 version = "0.1.0"
-source = "git+https://github.com/BreadchainCoop/commonware-restaking#07e5a4e9ebe26967b0f57bd83c0328058527e9a2"
+source = "git+https://github.com/BreadchainCoop/commonware-restaking#b20d0a512ae4e95c8e657b4b00403327d83fb86a"
 dependencies = [
  "alloy-primitives",
  "alloy-provider",
@@ -2230,7 +2230,7 @@ dependencies = [
 [[package]]
 name = "commonware-avs-node"
 version = "0.1.0"
-source = "git+https://github.com/BreadchainCoop/commonware-restaking#07e5a4e9ebe26967b0f57bd83c0328058527e9a2"
+source = "git+https://github.com/BreadchainCoop/commonware-restaking#b20d0a512ae4e95c8e657b4b00403327d83fb86a"
 dependencies = [
  "alloy",
  "alloy-network",
@@ -2260,7 +2260,6 @@ dependencies = [
  "futures",
  "futures-util",
  "governor",
- "prometheus-client 0.23.1",
  "prost",
  "prost-build",
  "rand 0.9.2",
@@ -2275,7 +2274,7 @@ dependencies = [
 [[package]]
 name = "commonware-avs-router"
 version = "0.1.0"
-source = "git+https://github.com/BreadchainCoop/commonware-restaking#07e5a4e9ebe26967b0f57bd83c0328058527e9a2"
+source = "git+https://github.com/BreadchainCoop/commonware-restaking#b20d0a512ae4e95c8e657b4b00403327d83fb86a"
 dependencies = [
  "alloy",
  "alloy-primitives",
@@ -2308,7 +2307,7 @@ dependencies = [
  "governor",
  "hex",
  "num-bigint 0.4.6",
- "prometheus-client 0.23.1",
+ "prometheus-client",
  "prost",
  "prost-build",
  "rand 0.9.2",
@@ -2395,7 +2394,7 @@ dependencies = [
  "num-integer",
  "num-rational",
  "num-traits",
- "prometheus-client 0.22.3",
+ "prometheus-client",
  "rand 0.8.5",
  "rand_distr",
  "thiserror 2.0.18",
@@ -2422,7 +2421,7 @@ dependencies = [
  "opentelemetry",
  "opentelemetry-otlp",
  "opentelemetry_sdk",
- "prometheus-client 0.22.3",
+ "prometheus-client",
  "rand 0.8.5",
  "rayon",
  "sha2 0.10.9",
@@ -3780,7 +3779,7 @@ dependencies = [
  "dotenv",
  "futures",
  "gas-analyzer-evmsketch",
- "prometheus-client 0.23.1",
+ "prometheus-client",
  "serde",
  "serde_json",
  "tokio",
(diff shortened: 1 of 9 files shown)


BreadrichEngels (approved):
Clean, well-scoped addition. Implementation is correct throughout.

A few things I checked:

**`dispatch_start` double-use is fine** — `Instant` is `Copy`, so `Option<Instant>` is also `Copy`. Both pattern-match sites get their own copy of the value; no borrow or move issue.

**Success-only gating is correct** — `round_latency_seconds` is observed inside `Ok(_)` only. Failed/timed-out rounds don't contribute, which is the right call since a receipt-timeout sample would inject artificial ~120 s outliers into the percentiles.

**Buckets look right** — block-time dominated, fine-grained at the low end (0.5–5 s), practical ceiling at 300 s. Consistent with `execution_duration_seconds`.

**Dashboard layout** — 4 × w=6 = 24, tiles cleanly. The section header update to mention ingress-queue wait is also excluded is a good clarification.

**Nit (not blocking):** `executor_chain_detection_seconds` is still a field on `MetricsCollector` but is no longer observed since #269 removed the executor-side chain detection. Worth a cleanup at some point, but unrelated to this PR.

Approve.

BreadrichEngels (commented):
Re-reviewed the new commit. The orchestrator-metrics wiring is clean and the expanded scope makes sense now that #169 is in.

Verified:
- **Wiring chain** — `main.rs` → `build(.., &context)` → `create_gas_killer_executor(.., context)` → `.with_metrics(context)` threads the runtime context to the BLS executor so `executor_state_retrieval_seconds` / `executor_operator_cache_misses_total` register on the registry that `/metrics` encodes. Correct.
- **`prometheus-client` 0.22.3 pin + `.into_iter()`** — consistent with the same fix in #169. 0.22's `Histogram::new` takes `impl Iterator<Item = f64>` (not `IntoIterator`), so the array literals need `.into_iter()`. All call sites updated.
- **Dashboard** — metric names match #169's scoped registration (`orchestrator_*`, `executor_*`), `status` label values line up (`Success`/`Invalid`/`Dropped`/`Failure`), and the layout tiles cleanly (4×w6 at y=33, 2×w12 at y=41, Ingress/Executor sections shifted to y=49/58). The "included in the P2P Round-Trip measurement" note on the state-retrieval panel is accurate — that read stretch runs between threshold and the contract handler, where `p2p_round_trip` is stamped.

---

**One thing worth a look before merge — unrelated transitive downgrades in `Cargo.lock`:**

```
proc-macro-crate 3.5.0 → 1.1.3
itertools 0.14.0 → 0.12.1   (prost-build, prost-derive)
```

These aren't pulled by `prometheus-client` (they're prost build-deps), so they look like incidental re-resolve churn rather than anything the 0.22.3 pin requires. A 2-major-version downgrade of `proc-macro-crate` in particular is worth confirming is intentional/acceptable rather than a side effect of regenerating the lockfile in a different environment. `check`/`clippy`/`test` passing suggests it's harmless, but tightening the lockfile diff to just the `prometheus-client` change would keep the dependency delta auditable.

Nothing blocking from the round-latency or dashboard code itself — just want a sanity check on that lockfile churn.

PR #273, opened 2026-06-12
feat: make P2P message backlog and rate limit configurable

### LINKED ISSUE #181: Make P2P message backlog and rate limit configurable
## Parent Epic
gas-killer/roadmap#9

## Context
`DEFAULT_MESSAGE_BACKLOG: usize = 256` is hardcoded in two places:
- `node/src/main.rs:347`
- `router/src/main.rs:272`

Both register the P2P channel with:
```rust
network.register(0, Quota::per_second(NZU32!(1)), DEFAULT_MESSAGE_BACKLOG)
```

There are two independent concerns here:

**1. Backlog size (256)** — the epic flags this as potentially too small under load. If the router is producing tasks faster than nodes can consume them, messages queue. Once the backlog fills, new messages are dropped silently. 256 is a reasonable default for low traffic but provides no headroom for bursts.

**2. Rate limit (`Quota::per_second(1)`)** — 1 message per second is extremely conservative. A single gas analysis takes on the order of seconds, so this may be the actual throughput ceiling, not the backlog. This needs measurement.

## Proposed Changes
- Replace `DEFAULT_MESSAGE_BACKLOG` with an env-var-backed config: `P2P_MESSAGE_BACKLOG` (default 256)
- Expose the rate quota as `P2P_MESSAGES_PER_SECOND` (default 1, accepts float for sub-second)
- Apply changes in both `node/src/main.rs` and `router/src/main.rs`
- Add the new env vars to `example.env` and the Helm chart values

## Files to Change
- `node/src/main.rs` (~line 347)
- `router/src/main.rs` (~line 272)
- `example.env`
- `helm/` values files

## Acceptance Criteria
- Both backlog and rate limit are configurable without recompiling
- Default values are unchanged (no behavior change without explicit config)
- New env vars are documented in `example.env` with comments explaining the tradeoff

## Unknowns / Risks
- Increasing the rate limit beyond what gas-analyzer can process causes queue buildup on the node side — needs load testing to find a safe ceiling
- `commonware-p2p v0.0.63` may have its own internal flow control that supersedes these values — needs verification

bagelface:
## Summary
- Adds `P2P_MESSAGE_BACKLOG` (default 256) and `P2P_MESSAGES_PER_SECOND` (default 1.0) env vars, replacing hardcoded values in `node/src/main.rs` and `router/src/main.rs`
- Returns 503 with a human-readable message at the ingress when `queue_depth >= P2P_MESSAGE_BACKLOG`, preventing silent drops at the P2P layer
- Adds `gas_killer_ingress_requests_at_capacity` Prometheus counter for observability
- Wires both vars into `example.env` and Helm chart values/templates

## Test plan
- [ ] Deploy with defaults — verify no behavior change
- [ ] Set `P2P_MESSAGE_BACKLOG=2` and send 3 rapid requests — third should receive 503 "Service at capacity"
- [ ] Set `P2P_MESSAGES_PER_SECOND=5.0` — verify throughput increase via `gas_killer_ingress_requests_accepted` metric
- [ ] Check `gas_killer_ingress_requests_at_capacity` increments correctly under load

Closes #181

### DIFF
--- a/common/src/config.rs
+++ b/common/src/config.rs
@@ -189,14 +189,27 @@ pub fn p2p_message_backlog() -> usize {
         .unwrap_or(DEFAULT_P2P_MESSAGE_BACKLOG)
 }
 
-/// Reads the P2P channel rate limit from `P2P_MESSAGES_PER_SECOND`, defaulting to
-/// [`DEFAULT_P2P_MESSAGES_PER_SECOND`].
-pub fn p2p_messages_per_second() -> f64 {
-    env::var("P2P_MESSAGES_PER_SECOND")
-        .ok()
-        .and_then(|v| v.parse().ok())
-        .filter(|&v: &f64| v > 0.0 && v.is_finite())
-        .unwrap_or(DEFAULT_P2P_MESSAGES_PER_SECOND)
+/// Reads the P2P channel rate limit from `P2P_MESSAGES_PER_SECOND` and returns the
+/// per-message quota period (`1 / rate`), defaulting to
+/// [`DEFAULT_P2P_MESSAGES_PER_SECOND`] when unset or invalid.
+///
+/// The quota is a smooth rate with no burst allowance: a rate of `5.0` permits one
+/// message every 200 ms, not bursts of five. Values whose reciprocal would overflow
+/// a `Duration` (e.g. `1e-20`) are treated as invalid and fall back to the default.
+pub fn p2p_quota_period() -> std::time::Duration {
+    parse_p2p_quota_period(env::var("P2P_MESSAGES_PER_SECOND").ok().as_deref())
+}
+
+/// Parses a `P2P_MESSAGES_PER_SECOND` value into a quota period, falling back to the
+/// default rate on malformed, non-positive, non-finite, or `Duration`-overflowing input.
+fn parse_p2p_quota_period(value: Option<&str>) -> std::time::Duration {
+    value
+        .and_then(|v| v.trim().parse::<f64>().ok())
+        .filter(|&v| v > 0.0 && v.is_finite())
+        .and_then(|v| std::time::Duration::try_from_secs_f64(1.0 / v).ok())
+        .unwrap_or_else(|| {
+            std::time::Duration::from_secs_f64(1.0 / DEFAULT_P2P_MESSAGES_PER_SECOND)
+        })
 }
 
 /// Maximum age (in blocks) of a reference block, falling back to the contract default
@@ -258,5 +271,45 @@ impl SpeculativePrebuildConfig {
             poll_interval: std::time::Duration::from_millis(poll_ms),
             confirmation_depth,
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::time::Duration;
+
+    #[test]
+    fn p2p_quota_period_default_is_one_per_second() {
+        assert_eq!(parse_p2p_quota_period(None), Duration::from_secs(1));
+    }
+
+    #[test]
+    fn p2p_quota_period_converts_rate_to_period() {
+        assert_eq!(
+            parse_p2p_quota_period(Some("5.0")),
+            Duration::from_millis(200)
+        );
+        assert_eq!(parse_p2p_quota_period(Some("0.5")), Duration::from_secs(2));
+    }
+
+    #[test]
+    fn p2p_quota_period_rejects_invalid_values() {
+        let default = Duration::from_secs(1);
+        assert_eq!(parse_p2p_quota_period(Some("")), default);
+        assert_eq!(parse_p2p_quota_period(Some("abc")), default);
+        assert_eq!(parse_p2p_quota_period(Some("0")), default);
+        assert_eq!(parse_p2p_quota_period(Some("-1.5")), default);
+        assert_eq!(parse_p2p_quota_period(Some("inf")), default);
+        assert_eq!(parse_p2p_quota_period(Some("NaN")), default);
+    }
+
+    #[test]
+    fn p2p_quota_period_rejects_duration_overflow() {
+        // 1.0 / 1e-20 overflows Duration; must fall back to the default, not panic.
+        assert_eq!(
+            parse_p2p_quota_period(Some("1e-20")),
+            Duration::from_secs(1)
+        );
     }
 }
--- a/common/src/config.rs
+++ b/common/src/config.rs
@@ -195,18 +195,21 @@ pub fn p2p_message_backlog() -> usize {
 ///
 /// The quota is a smooth rate with no burst allowance: a rate of `5.0` permits one
 /// message every 200 ms, not bursts of five. Values whose reciprocal would overflow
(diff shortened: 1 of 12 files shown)


BreadrichEngels (commented):
Solid PR — the config plumbing is clean, the env-var validation is careful, and the 503-at-capacity response is a real improvement over silent P2P-layer drops. One design point worth discussing before merge, plus a couple of minor notes.

---

**1. Capacity check runs *after* `validate_onchain` — load-shedding happens too late**

In `trigger_task_handler` the order is:

1. bearer auth
2. `request.validate()` (cheap)
3. `validate_onchain(...)` — **multiple RPC round-trips** (`eth_getCode`, block number, `stateTransitionCount`)
4. capacity check → 503
5. enqueue

The problem: the capacity check is the thing that protects the service when it's overloaded, but it only fires *after* every rejected request has already paid the full on-chain validation cost. So precisely when the queue is saturated — i.e. the system is already struggling — each incoming request still hammers the RPC with validation calls before getting its 503. Since RPC time is the documented bottleneck here, that's a feedback loop that makes overload worse rather than shedding it.

I'd move the capacity check to before `validate_onchain` (it only reads an atomic, so it's essentially free). The tradeoff is that a malformed request arriving during overload would get a 503 instead of a 400 — but it was going to fail anyway, so early load-shedding is the better behavior. The soft-limit race (concurrent requests all passing the check before any enqueues) is unchanged by moving it earlier.

---

**2. Quota semantics changed from burst-capable to smooth (probably intended, worth confirming)**

```rust
Quota::with_period(Duration::from_secs_f64(1.0 / p2p_messages_per_second()))
```

`Quota::with_period` has a max burst of 1, whereas the old `Quota::per_second(NZU32!(1))` was also burst-1 — so default behavior is preserved (good, matches test-plan item 1). But note that at higher rates this stays burst-1: `P2P_MESSAGES_PER_SECOND=5.0` becomes "one message every 200ms," **not** "a burst of 5 then refill." That's arguably the safer choice for protecting gas-analyzer throughput, just want to confirm it's the intended semantics since the env-var name reads like it'd allow a 5-message burst.

---

**3. Minor: `Duration::from_secs_f64` panic path**

`p2p_messages_per_second()` filters `> 0.0 && is_finite()`, which is great, but an absurdly tiny value (e.g. `1e-20`) passes the filter and makes `1.0 / rate` overflow `Duration`, panicking inside `from_secs_f64` with a message that won't match the `.expect("...must be positive and finite")`. Not worth special handling, but a sane lower bound (or just noting it) would make the failure mode match the message.

---

**Minor / non-blocking:**
- No Grafana panel for the new `gas_killer_ingress_requests_at_capacity` counter — worth a follow-up so the 503s are visible alongside the existing Ingress panels.
- The two capacity tests use `without_metrics`, so the `ingress_at_capacity.inc()` branch isn't exercised. Trivial, but a metrics-present variant would close that gap.

The boundary tests (`test_queue_one_below_limit_still_accepts` at depth=1/max=2, `test_full_queue_returns_503` at depth=1/max=1) are well chosen, and quoting `p2pMessagesPerSecond: "1.0"` as a string in values.yaml to avoid YAML float normalization is a nice touch.

bagelface:
Addressed in c3ea865 — thanks for the thorough review.

**1. Capacity check moved before validation.** It now runs immediately after bearer auth, before both `request.validate()` and `validate_onchain()`, so an overloaded service sheds load without paying any RPC round-trips. Took your suggested tradeoff as-is: malformed requests during overload get a 503 instead of a 400.

**2. Burst-1 semantics confirmed intended.** The smooth rate is deliberate — the limit exists to protect gas-analyzer throughput, and a burst allowance would just let bursts queue behind multi-second analyses. Documented in `example.env` and the `p2p_quota_period` doc comment ("5.0 means one message every 200ms, not bursts of five") so the env-var name doesn't mislead.

**3. Overflow panic path eliminated.** The rate→period conversion moved into the config layer as `p2p_quota_period()`, which uses `Duration::try_from_secs_f64` — values like `1e-20` now fall back to the default instead of panicking with a mismatched expect message. Covered by a unit test (`p2p_quota_period_rejects_duration_overflow`) along with the other parse edge cases.

**Minor items:** added an "at capacity (503)" series to the existing Ingress Request Rate panel, and a metrics-present test (`test_full_queue_increments_at_capacity_metric`) asserting `ingress_at_capacity` increments while `ingress_rejected` stays at 0.

BreadrichEngels (approved):
Re-review of `c3ea865`. All three points from the last pass are addressed, verified in-code — nice work.

**1. Load-shedding moved ahead of validation ✅**
`trigger_task_handler` now runs the capacity check immediately after bearer auth, before both `request.validate()` and `validate_onchain()`. An overloaded service now sheds load without paying any RPC round-trips, which was the whole point. The early-503-instead-of-400 tradeoff for malformed requests during overload is the right call. The soft-limit race is unchanged as expected. Bonus: wiring `max_queue_depth` from `p2p_message_backlog()` in `factories.rs` means the ingress rejection threshold matches the actual P2P channel backlog, so the 503 fires right at the point the channel would otherwise drop silently — exactly the consistency you want.

**2. Burst-1 / smooth-rate semantics ✅**
Confirmed intended and now documented in both `example.env` and the `p2p_quota_period` doc comment ("5.0 means one message every 200ms, not bursts of five"), so the env-var name no longer misleads.

**3. Overflow panic eliminated ✅**
Moving the rate→period conversion into `parse_p2p_quota_period` with `Duration::try_from_secs_f64` + default fallback is clean, and pulling it into a free function gave you the nice table of unit tests (`""`, `abc`, `0`, `-1.5`, `inf`, `NaN`, `1e-20`). Good coverage.

**Minor items ✅** — "at capacity (503)" series added to the Ingress panel, and `test_full_queue_increments_at_capacity_metric` now exercises the metrics-present branch (asserting `ingress_at_capacity == 1` while `ingress_rejected == 0`).

---

**One residual edge (non-blocking), symmetric to the overflow you just fixed:**

You closed the *tiny-rate → period overflow* hole, but the *huge-rate → period underflows to zero* hole is still open. For a very large `P2P_MESSAGES_PER_SECOND` (above ~2e9), `1.0 / rate` rounds below `Duration`'s 1ns resolution, so `try_from_secs_f64` returns `Ok(Duration::ZERO)` (zero is valid, not an error). That zero period then makes `Quota::with_period(..)` return `None`, and the `.expect("p2p_quota_period always returns a non-zero duration")` in `main.rs` panics at startup — so the invariant the expect message asserts doesn't strictly hold.

It's an absurd config value and startup-only, so I'm not blocking on it, but since you're already validating this input it'd be tidy to either clamp to a sane minimum period or add a `> 0` (non-zero `Duration`) filter in `parse_p2p_quota_period` so the `.expect()` invariant is actually guaranteed. A `p2p_quota_period_rejects_excessive_rate` test alongside the overflow one would mirror it.

Approving — feel free to fold the above into a follow-up rather than hold the merge. Heads-up that no CI checks are currently reporting on the head commit (status came back empty), so make sure the test job actually ran before merging.

Issue #276, opened 2026-06-24
Remove startup stagger workaround once operator state cache is implemented


bagelface:
## Context

#275 tracks a disk cache for operator states to eliminate repeated \`eth_getLogs\` sweeps on startup. Once that is implemented with a shared cache file and read-before-scan coordination, the startup stagger added across both docker-compose and Helm is no longer needed.

## What to remove

**docker-compose.yml** — revert node-2, node-3, and router back to their original entrypoint/command form:

- **node-2**: remove \`entrypoint\` override and \`sleep 15\`
- **node-3**: remove \`entrypoint\` override and \`sleep 30\`
- **router**: remove \`entrypoint\` override and \`sleep 45\`

**Helm charts** — remove the \`startup-delay\` init container from both templates and the driving value:

- \`helm/gas-killer/templates/node-deployment.yaml\`: remove the \`startup-delay\` init container block (conditioned on \`node.startupDelaySecsPerNode > 0 && nodeIndex > 1\`)
- \`helm/gas-killer/templates/router-deployment.yaml\`: remove the \`startup-delay\` init container block (conditioned on \`nodeCount * startupDelaySecsPerNode > 0\`)
- \`helm/gas-killer/values.yaml\`: remove \`node.startupDelaySecsPerNode\`

## Known cost of the Helm approach

Unlike docker-compose where the sleep is a one-time startup cost, the init container runs on **every pod (re)start** in K8s — crash loops, rescheduling, rolling updates. A crash-looping node-3 eats 30s before rejoining each time. Acceptable as a stopgap since it disappears with #275, but worth knowing the per-restart overhead exists in prod.

## Condition

This should only be done after #275 is merged **and** the cache implementation handles the first cold-start case (i.e. the shared file + read-before-scan pattern so only one process runs a full sweep when no cache exists yet). If the cache is implemented naively (per-process, no coordination), the stagger for cold starts should be kept.

## Related

- #275 — operator state disk cache (prerequisite)
- BreadchainCoop/commonware-restaking#177 — upstream chunk size fix

bagelface:
Removed as part of gas-killer/roadmap#19 #396, which implements the operator-state cache (#275) and then sets `startupDelaySecsPerNode: 0`.

[...]

Issue #275, opened 2026-06-24
Cache operator states to disk to eliminate repeated eth_getLogs sweeps on startup


bagelface:
## Problem

Every call to `get_operator_states()` (`common/src/config.rs:158`) triggers a full `eth_getLogs` backfill via the EigenSDK from the contract deploy block to the current chain head. This runs two parallel sweeps:

- **BLS pub key sweep** (`eigen-client-avsregistry`): 1,024-block chunks → `eth_getLogs` per chunk
- **Operator socket sweep**: 10,000-block chunks → `eth_getLogs` per chunk

With a ~68k-block scan range (~74 `eth_getLogs` calls at 75 CU each), each sweep costs ~5,664 Alchemy CU. With 4 processes (router + 3 nodes), this caused a spike of **~12,010 CU/s** on startup. Even after staggering (PR #TODO), every cold start and every `restart: unless-stopped` recovery incurs the full cost.

## Proposed solution

After a successful scan, persist the operator state and the scanned-to block to a JSON cache file alongside `avs_deploy.json` (e.g. `operator_state_cache.json`):

```json
{
  "scanned_to_block": 11050000,
  "operators": [ ... ]
}
```

On the next startup, load the cache and only scan the delta from `scanned_to_block + 1` to the current block. Since operators rarely register mid-run, the delta scan is typically 0–2 `eth_getLogs` calls, reducing per-restart cost from ~5,664 CU to near zero.

**Cache invalidation**: the cache is always valid to load and delta-scan forward from — no new registrations are missed since the delta covers any blocks added while the service was down. The cache can be keyed by `registry_coordinator_address` to avoid cross-deployment pollution.

## Implementation notes

- Lives in `common/src/config.rs` around `get_operator_states()`
- Cache path derived from `AVS_DEPLOYMENT_PATH` (same directory, fixed filename)
- Must handle missing/corrupt cache gracefully (fall back to full scan)
- For option 2 from the startup CU analysis: if the cache file is written by the first process to complete, subsequent processes started immediately after can read from it — eliminating the need for them to scan at all

## Expected impact

~100% reduction in `eth_getLogs` CU on all restarts after the first cold start. Also reduces the need for startup staggering as a workaround.

## Related

- Startup stagger (docker-compose): #TODO
- Upstream chunk size fix (commonware-restaking): BreadchainCoop/commonware-restaking#TODO

bagelface:
[...]

PR #277, opened 2026-06-24
fix: reduce eth_getLogs RPC hammering on startup

bagelface:
## Summary

- Eliminates a redundant `get_operator_states()` call in the node — the second call was re-running a full `eth_getLogs` backfill for data already loaded earlier in the same startup sequence
- Staggers Docker Compose startup so at most one process runs its EigenSDK log sweep at a time: node-1 at T=0s, node-2 at T=15s, node-3 at T=30s, router at T=45s

## Background

Each `get_operator_states()` call triggers two parallel `eth_getLogs` sweeps via the EigenSDK (BLS pub key sweep in 1,024-block chunks, socket sweep in 10,000-block chunks). With a ~68k-block scan range this costs ~5,664 Alchemy CU per call. Before this fix, 7 calls ran concurrently at startup (router × 1, each node × 2), producing a spike of ~12,010 CU/s.

**After this PR: ~1,715 CU/s** (4 calls, serialized).

## Follow-up

- #275 — operator state disk cache (eliminates the sweep entirely on restarts, and unblocks removal of the stagger)
- #276 — remove the startup stagger once #275 is implemented
- BreadchainCoop/commonware-restaking#177 — upstream chunk size fix (1,024 → 10,000 blocks for pub key sweep, ~5× fewer calls)

## Test plan

- [ ] `docker compose up` — confirm services start in order with ~15s gaps between node-2, node-3, router
- [ ] Verify Alchemy CU/s no longer spikes above ~2,000 on startup
- [ ] Confirm all nodes and router successfully connect and process a task end-to-end

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -97,15 +97,9 @@ services:
       - ./config/.nodes/operator_keys/testacc2.private.bls.key.json:/app/key.json:ro
       - ./config/.nodes:/app/.nodes:ro
       - ./config/public_orchestrator.dev.json:/app/public_orchestrator.json:ro
+    entrypoint: ["/bin/sh", "-c"]
     command:
-      [
-        "--key-file",
-        "/app/key.json",
-        "--port",
-        "3002",
-        "--orchestrator",
-        "/app/public_orchestrator.json",
-      ]
+      - "sleep 15 && exec gas-killer-node --key-file /app/key.json --port 3002 --orchestrator /app/public_orchestrator.json"
     ports:
       - "4002:3002"
       - "9002:8081"  # metrics / healthz
@@ -131,15 +125,9 @@ services:
       - ./config/.nodes/operator_keys/testacc3.private.bls.key.json:/app/key.json:ro
       - ./config/.nodes:/app/.nodes:ro
       - ./config/public_orchestrator.dev.json:/app/public_orchestrator.json:ro
+    entrypoint: ["/bin/sh", "-c"]
     command:
-      [
-        "--key-file",
-        "/app/key.json",
-        "--port",
-        "3003",
-        "--orchestrator",
-        "/app/public_orchestrator.json",
-      ]
+      - "sleep 30 && exec gas-killer-node --key-file /app/key.json --port 3003 --orchestrator /app/public_orchestrator.json"
     ports:
       - "4003:3003"
       - "9003:8081"  # metrics / healthz
@@ -170,7 +158,9 @@ services:
     volumes:
       - ./config/.nodes:/app/.nodes:ro
       - ./config/router_orchestrator.dev.json:/app/router_orchestrator.json:ro
-    command: ["--key-file", "/app/router_orchestrator.json", "--port", "3000"]
+    entrypoint: ["/bin/sh", "-c"]
+    command:
+      - "sleep 45 && exec gas-killer-router --key-file /app/router_orchestrator.json --port 3000"
     ports:
       - "4000:3000"
       - "8080:8080"
--- a/helm/gas-killer/templates/node-deployment.yaml
+++ b/helm/gas-killer/templates/node-deployment.yaml
@@ -154,6 +154,19 @@ spec:
               echo "L1 is ready!"
         {{- end }}
 
+        {{- $delayPerNode := int $root.Values.node.startupDelaySecsPerNode }}
+        {{- $nodeDelay := mul $i $delayPerNode }}
+        {{- if gt $nodeDelay 0 }}
+        # Stagger EigenSDK eth_getLogs backfill: node-{{ $nodeIndex }} waits {{ $nodeDelay }}s so at most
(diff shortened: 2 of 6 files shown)


BreadrichEngels (approved):
Both changes check out — verified each rather than eyeballing. LGTM.

**Node dedup (`node/src/main.rs`) — correct.** I pulled the full file: `quorum_infos` is declared before the "avoid configuring two loggers" block, assigned inside it, and only *borrowed* there (`quorum_infos.len()`, `quorum_infos[quorum_number].operators.clone()` — a clone, not a move). So reusing it after the block (`let operators = &quorum_infos[quorum_number].operators;` and `.threshold`) is sound, and the deferred `let quorum_infos;` is definitely-assigned (the block unconditionally assigns or panics via `.expect`). Dropping the second `get_operator_states()` is not just a CU win — it also removes a small TOCTOU window, since peer registration and the contributor/G1 list now derive from one snapshot instead of two separate sweeps.

**Compose stagger — correct.** Confirmed both images use a bare-binary entrypoint (`ENTRYPOINT ["gas-killer-node"]` / `["gas-killer-router"]`), so `exec gas-killer-node …` / `exec gas-killer-router …` resolve on PATH — no absolute-path breakage from overriding `entrypoint` to `/bin/sh -c`. `exec` keeps the binary as PID 1 so SIGTERM still propagates (important for clean shutdown), and the `/bin/sh -c` pattern is already used by the anvil service in this same file, so it's consistent. node-1 is left at T=0, matching the described 0/15/30/45s ladder. Arg lists match the originals.

**Notes (non-blocking):**
- **Scope of the win.** The node dedup is compiled into the binary, so it helps *every* deployment including prod Helm (one sweep per node instead of two). The 15/30/45s stagger is `docker-compose`-only — prod pod startup is scheduled by K8s, not this file — so the headline ~12,010→~1,715 CU/s figure is the local-dev number. Worth keeping in mind that the real prod startup-sweep fix is the #275 disk cache; this PR's prod benefit is the halved sweep count, not the serialization.
- **Shell availability.** The stagger assumes `/bin/sh` + `sleep` exist in the node/router runtime images. They do if those images are debian/alpine-based (and anvil already shells out here), but this would break on a distroless runtime — your `docker compose up` test-plan item will catch it either way.
- **Local e2e latency.** Router now starts 45s later, so local end-to-end spin-up is slower until #276 removes the stagger — expected tradeoff, just flagging for anyone timing local runs.
- CI is `unstable` with no checks reporting on the head commit yet — confirm green before merge.

Nice, well-scoped fix with the follow-ups (#275/#276 and the upstream chunk-size fix #177) already mapped out. Approving.

BreadrichEngels (approved):
Re-review — the new revision extends the stagger to prod via Helm and hardens the e2e script, which addresses the two notes from my last pass (prod wasn't covered; the 45s delay would confuse the old health check). All verified.

**Helm stagger (node + router) — correct.** I read `node-deployment.yaml`: the loop is `range $i := until $nodeCount`, so `$i` is 0-based and `$nodeDelay = mul $i startupDelaySecsPerNode` gives node-1=0 (skipped by `gt $nodeDelay 0`), node-2=15s, node-3=30s — the intended ladder. The router's `mul (int .Values.global.nodeCount) (int .Values.node.startupDelaySecsPerNode)` = 45s for 3 nodes, and crucially it's keyed off the same `global.nodeCount` that drives the node loop, so the node and router delays can't drift apart. The `startup-delay` initContainer is appended *after* wait-for-setup/bridge/L1, so the sleep happens immediately before the main container (i.e. right before the `eth_getLogs` sweep), which is exactly where you want it. `busybox:1.36` matches the image the other initContainers already use, so sh+sleep are present — no distroless risk, which was my earlier shell-availability concern for the prod path.

**e2e health poll — hits a real route.** Confirmed the router serves `/healthz` on the ingress server, not just the K8s healthz port: `router/src/ingress.rs`'s `build_app()` registers `.route("/healthz", get(healthz_handler))`, and compose maps the ingress `8080:8080`. So `curl http://localhost:8080/healthz` returns 200 and the new wait loop is valid. The 120s timeout comfortably covers the 45s router stagger + sweep, and dumping `docker compose logs --tail=50 router` on timeout is a nice touch for debugging. This is strictly better than the old `docker compose ps | grep Up` check, which only confirmed the container was up, not that ingress was actually serving.

**values.yaml** — `startupDelaySecsPerNode: 15` is documented with the disable (`0`) and removal (#275) notes. Good.

**One new minor (non-blocking):** the initContainer delay runs on *every* pod (re)start, not just first boot — so in prod a crash-looping or rescheduled node-3 eats its 30s before rejoining each time, unlike the one-shot compose case. Acceptable for a stopgap and it disappears with the #275 disk cache, just worth being aware it's a per-restart cost in K8s, not a one-time one.

Node dedup + compose stagger are unchanged from my prior approval and still good. CI still isn't reporting on the head commit — get it green before merge. Re-approving.

PR #279, opened 2026-06-24
feat: update to "commonware-restaking" orchestrator latency improvements

bagelface:
## Summary

Updates the service to be compatible with the breaking changes introduced by [BreadchainCoop/commonware-restaking#175](https://github.com/BreadchainCoop/commonware-restaking/pull/175), bumping the dependency to the current `main` tip.

- **`Creator` trait**: implements the new required `wait_for_new_round(current: u64)` method on all three creator types (`GasKillerCreator`, `ListeningGasKillerCreator`, `GasKillerCreatorType`)
- **Env var rename**: `AGGREGATION_TIMEOUT` → `ROUND_TIMEOUT` + new `REBROADCAST_INTERVAL` (updated in `example.env`, Helm values, deployment template, and Grafana dashboard copy)
- **Cargo.lock**: bumps all `commonware-avs-*` crates from `623d4231` → `8f3c3277`

## Test plan

- [ ] `cargo check --all-targets --all-features` passes
- [ ] `cargo clippy --all-targets --all-features -- -D warnings` passes (zero warnings)
- [ ] `cargo test --all-targets --all-features` passes (97 tests, 0 failures)
- [ ] Helm: `ROUND_TIMEOUT` and `REBROADCAST_INTERVAL` injected correctly into router pod via `values.yaml` → `router-deployment.yaml`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2175,7 +2175,7 @@ checksum = "1d07550c9036bf2ae0c684c4297d503f838287c83c53686d05370d0e139ae570"
 [[package]]
 name = "commonware-avs-bindings"
 version = "0.1.0"
-source = "git+https://github.com/BreadchainCoop/commonware-restaking#623d4231123dfa7853c08d6e6ec8c624d64dc9b3"
+source = "git+https://github.com/BreadchainCoop/commonware-restaking#8f3c32773dfa07af8b8f25cc6f445a79bde2ec4f"
 dependencies = [
  "alloy",
  "alloy-primitives",
@@ -2190,7 +2190,7 @@ dependencies = [
 [[package]]
 name = "commonware-avs-core"
 version = "0.1.0"
-source = "git+https://github.com/BreadchainCoop/commonware-restaking#623d4231123dfa7853c08d6e6ec8c624d64dc9b3"
+source = "git+https://github.com/BreadchainCoop/commonware-restaking#8f3c32773dfa07af8b8f25cc6f445a79bde2ec4f"
 dependencies = [
  "alloy",
  "anyhow",
@@ -2212,7 +2212,7 @@ dependencies = [
 [[package]]
 name = "commonware-avs-eigenlayer"
 version = "0.1.0"
-source = "git+https://github.com/BreadchainCoop/commonware-restaking#623d4231123dfa7853c08d6e6ec8c624d64dc9b3"
+source = "git+https://github.com/BreadchainCoop/commonware-restaking#8f3c32773dfa07af8b8f25cc6f445a79bde2ec4f"
 dependencies = [
  "alloy-primitives",
  "alloy-provider",
@@ -2230,7 +2230,7 @@ dependencies = [
 [[package]]
 name = "commonware-avs-node"
 version = "0.1.0"
-source = "git+https://github.com/BreadchainCoop/commonware-restaking#623d4231123dfa7853c08d6e6ec8c624d64dc9b3"
+source = "git+https://github.com/BreadchainCoop/commonware-restaking#8f3c32773dfa07af8b8f25cc6f445a79bde2ec4f"
 dependencies = [
  "alloy",
  "alloy-network",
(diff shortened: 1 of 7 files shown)


BreadrichEngels (approved):
This is the lockstep PR I've been flagging across the #172/#173/#174 → #175 chain, and it lands all three coordination items correctly. I verified the two that actually matter rather than trusting the summary.

**1. The `Cargo.lock` bump points at the right commit.** `8f3c3277` is exactly the merge commit of BreadchainCoop/commonware-restaking#175 ("Merge pull request #175 … orchestrator-refactor"), so the dependency genuinely contains the renamed env reading + `wait_for_new_round` + the decoupled/provider timing. ✅

**2. The production creator is safe from the double-consume bug I flagged on #172.** This was my main worry, and it's handled — `ListeningGasKillerCreator`'s `round` is a per-task monotonic `round_counter` (`fetch_add(1)`), so after the orchestrator executes round R, the next `get_payload_and_round()` always returns R+1 > R. The `wait_for_new_round` loop therefore returns on its first iteration, consuming exactly one task — no every-other-task drop, and no double EVMSketch compute (the orchestrator uses the cached `(payload, round)` rather than re-fetching). That's precisely the single-consume contract the merged orchestrator relies on. The `if result.1 > current` guard is effectively always-true defensive code here, which is fine. (The `Basic` creator's increment-and-sleep impl is fine too — it's the non-ingress dev path.)

**3. The env rename carries the 300 across — no silent revert.** `roundTimeout: "300"` + `rebroadcastInterval: "300"` in values.yaml, wired into the router pod, plus example.env, the dashboard copy, and the values.yaml grace-period comments. Setting `REBROADCAST_INTERVAL == ROUND_TIMEOUT == 300` means the orchestrator's biased select never fires an intra-round rebroadcast (timeout wins the tie), so this reproduces the prior single-knob 300s behaviour exactly — the right call, and it closes the silent-30s-fallback risk from the #173 discussion.

---

**One thing this PR misses (non-blocking but should be a quick follow-up): `README.md` still documents `AGGREGATION_TIMEOUT`.** The four other locations are all updated, but `README.md` (line ~106, the "Optional environment variables" list) isn't in the diff and still describes `AGGREGATION_TIMEOUT` — which no longer exists after this merges. Please rename it to `ROUND_TIMEOUT` and add `REBROADCAST_INTERVAL` so the docs match the deployed knobs.

**Minor:**
- **`Cargo.lock` transitive flip-flop.** `proc-macro-crate` (3.5.0→1.1.3) and `itertools` (0.14.0→0.12.1) revert to their pre-#274 versions — the mirror image of the bump #274 introduced. Harmless if it builds, but the back-and-forth suggests the lock is being regenerated in differing environments; worth confirming `cargo check`/CI is clean so this isn't a hand-edited or stale resolution.
- **Comment precision.** The example.env / values.yaml note says rebroadcast is disabled by setting the interval *longer than* `ROUND_TIMEOUT` — but the shipped default has them *equal* (300/300), which also disables it via the biased tie. Tiny wording nit; the behaviour is correct.
- No dedicated test for the listening creator's `wait_for_new_round` (one-task-per-round) — the contract is covered upstream by the orchestrator tests and holds by the monotonic counter, so optional, but a small test asserting `round` advances by exactly one per call would lock it in.

**Gating:** `mergeable_state` is `unstable` and no checks are reporting yet. This PR is compile-critical — the three `Creator` impls have to build against the new dependency — so the `cargo check`/`clippy`/`test` run is the real gate here; don't merge until it's green. And per the chain, this should merge + deploy together with the #175 side already on `main` (it is). Approving on the code; just land the README fix and confirm CI.

BreadrichEngels (approved):
Re-review — all three follow-ups from my last pass are in. 

- **README** now documents `ROUND_TIMEOUT` and adds the `REBROADCAST_INTERVAL` entry, so the env-var docs match the deployed knobs — that was the last stale `AGGREGATION_TIMEOUT` reference, and a search confirms nothing else in the repo still uses the old name. ✅
- **Comment precision** fixed in both `example.env` and `values.yaml` ("when equal to or longer than `ROUND_TIMEOUT` the biased select ensures no intra-round rebroadcast fires") — accurately describes the shipped 300/300 default. ✅
- **Test added** — `test_wait_for_new_round_strictly_advances` asserts the round advances on each call (and `start_paused` keeps the Basic creator's 2s sleep from slowing the suite). Covers the strictly-advancing contract the orchestrator relies on. ✅

The substance I verified last time is unchanged and still correct: lock pinned to `8f3c3277` (the #175 merge commit), and the production `ListeningGasKillerCreator` consumes exactly one task per round via the monotonic `round_counter`. 

The only remaining item is the one I couldn't sign off from outside CI: `mergeable_state` is still `unstable` with no checks reporting. This PR is compile-critical (the three `Creator` impls must build against the bumped dependency) and the `Cargo.lock` transitive flip-flop needs a clean resolve, so confirm `cargo check`/`clippy`/`test` are green before merging. Code's good — approving.

PR #282, opened 2026-06-25
feat: make inter-round poll interval configurable via POLLING_INTERVAL_MS

bagelface:
## Summary
- Adds a `polling_interval: Duration` field to `GasKillerCreator`, replacing the hardcoded 2-second sleep in `wait_for_new_round`
- `GasKillerCreator::new()` reads `POLLING_INTERVAL_MS` from the environment (default: 2000ms), consistent with how other config is loaded in the same file
- Documents `POLLING_INTERVAL_MS` in `example.env`

## Test plan
- [ ] Existing behaviour unchanged when `POLLING_INTERVAL_MS` is unset (defaults to 2000ms)
- [ ] Setting `POLLING_INTERVAL_MS` to a different value changes the poll cadence

### DIFF
--- a/example.env
+++ b/example.env
@@ -163,6 +163,9 @@ SIGNER_ENDPOINT=http://signer:50051
 INGRESS=true
 INGRESS_ADDRESS=0.0.0.0:8080
 INGRESS_TIMEOUT_MS=0
+
+# How often (in milliseconds) to poll for a new round between task cycles.
+POLLING_INTERVAL_MS=2000
 # INGRESS_PASSWORD=  # Required for TESTNET mode
 
 # =============================================================================
--- a/example.env
+++ b/example.env
@@ -165,6 +165,8 @@ INGRESS_ADDRESS=0.0.0.0:8080
 INGRESS_TIMEOUT_MS=0
 
 # How often (in milliseconds) to poll for a new round between task cycles.
+# Only applies when INGRESS=false (Basic creator). When INGRESS=true the creator
+# blocks on an event-driven channel and this value has no effect.
 POLLING_INTERVAL_MS=2000
 # INGRESS_PASSWORD=  # Required for TESTNET mode
 
--- a/example.env
+++ b/example.env
@@ -56,6 +56,11 @@ FORK_URL=https://ethereum-sepolia-rpc.publicnode.com
 # or longer than ROUND_TIMEOUT the biased select ensures no intra-round rebroadcast fires.
 # REBROADCAST_INTERVAL=300
 
+# How often (in milliseconds) to poll for a new round between task cycles (default: 2000).
+# Only applies when INGRESS=false; when INGRESS=true the creator blocks on an event-driven
+# channel and this value has no effect. Must be > 0 to avoid a busy-loop.
+# POLLING_INTERVAL_MS=2000
+
 # =============================================================================
 # P2P Channel Configuration
 # =============================================================================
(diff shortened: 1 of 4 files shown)


BreadrichEngels (commented):
The code change itself is correct and clean — `GasKillerCreator` gains a `polling_interval` field, `new()` reads `POLLING_INTERVAL_MS` (default 2000ms, consistent with how `GasKillerConfig` reads `INGRESS_TIMEOUT_MS` in the same file), the manual `Default` delegates to `new()`, and `wait_for_new_round` uses it instead of the hardcoded 2s. No signature break. That part is fine.

But there's a wiring problem I want to flag before this merges, because as-is the new Helm knob does nothing in production:

**🚩 `POLLING_INTERVAL_MS` only affects `GasKillerCreator` (Basic), but the deployed router runs `ListeningGasKillerCreator` (ingress).** The router runs with `INGRESS=true` (see example.env, and it's the whole point of the router), so it uses `ListeningGasKillerCreator`. That creator's `wait_for_new_round` is event-driven — it loops on `get_payload_and_round()` → `wait_for_task()`, which *blocks on the task channel* and returns as soon as a task arrives (the round is a monotonic per-task counter, so it returns on the first iteration). There is no poll-interval sleep on that path. Only `GasKillerCreator::wait_for_new_round` (the non-ingress Basic creator) sleeps on `polling_interval`.

So injecting `POLLING_INTERVAL_MS` / `pollingIntervalMs` into the *router* pod (`router-deployment.yaml` + `values.yaml`) is a no-op for the deployed config — an operator who tunes `router.pollingIntervalMs` expecting to change inter-round cadence will see no effect, since the ingress router never reads it on a hot path. Two ways to make this honest:

1. **Drop the Helm wiring** (router-deployment + values.yaml) and keep only the Rust + example.env change, since the Basic creator isn't what the router deployment runs. This is the cleaner option if the knob is really just for the Basic/non-ingress path.
2. **Keep it but document the scope** — annotate `pollingIntervalMs` in values.yaml as "only applies in non-ingress (Basic creator) mode; no effect when INGRESS=true," so it doesn't read as a production tunable it isn't.

If the actual goal was to tune the *ingress* router's behaviour, then this knob is on the wrong creator entirely and the Listening path has nothing to tune (it's blocking, not polling) — worth confirming what you intended.

**Minor:**
- Same `POLLING_INTERVAL_MS=0` → busy-loop foot-gun as the library side (only matters for the Basic creator, and only if explicitly set to 0).
- example.env placement inserts the new var + comment between `INGRESS_TIMEOUT_MS=0` and the `# INGRESS_PASSWORD=` comment, slightly splitting that group — trivial, just tidy if you touch it again.

For cross-reference: this parallels BreadchainCoop/commonware-restaking#182 (same knob in the example `CounterCreator`), but the two are independent — the service doesn't depend on that example crate, so no merge-order coupling and no Cargo.lock bump needed here.

Leaving this as a comment rather than approve/block — the Rust change is sound, but I'd like you to resolve the Helm no-op (drop it or scope-document it) so the deployed config isn't misleading. Happy to re-review once that's settled. CI is also `unstable`/not reporting yet — confirm green before merge.

BreadrichEngels (approved):
Re-review — all three points from my last pass are resolved. 

- **Helm no-op now documented (option 2).** Both `values.yaml` and `example.env` spell out that `pollingIntervalMs`/`POLLING_INTERVAL_MS` "only applies when INGRESS=false (Basic creator); when INGRESS=true the creator blocks on an event-driven channel and this value has no effect." That's exactly what I was after — keeping the knob is fine now that it no longer reads as a production tunable it isn't. ✅
- **`=0` busy-loop closed.** `new()` now filters `.filter(|&ms| ms > 0)` before `.unwrap_or(2_000)`, so `POLLING_INTERVAL_MS=0` (and unparseable values) fall back to the 2000ms default instead of turning `wait_for_new_round` into a no-sleep loop — and the example.env comment documents the constraint. You went a step past my note and actually enforced it. ✅
- **example.env placement** moved up next to the round-timing config, so it no longer splits the `INGRESS_TIMEOUT_MS` / `INGRESS_PASSWORD` group. ✅

The core Rust change remains correct: `polling_interval` field, env-read in `new()` (consistent with `INGRESS_TIMEOUT_MS` handling), manual `Default` delegating to `new()`, no signature break. Good to merge once CI is green — `mergeable_state` is now `clean` but no checks are reporting yet, so confirm `fmt`/`clippy`/`test` pass first. Approving.

(As before: independent of commonware-restaking#182 — same knob, different creator, no dependency coupling.)

Issue #122, opened 2026-03-06
Round counter resets on router restart, causing nodes to reject duplicate rounds


bagelface:
## Problem

`ListeningGasKillerCreator` uses an `AtomicU64` round counter (`creator.rs:184`) that starts at 0 on every pod restart. If a node has already signed round 0 and the router restarts, it will propose round 0 again and nodes will refuse to sign (they track signed rounds in a `HashSet` and skip duplicates).

This means any router restart effectively stalls the system until nodes are also restarted.

## Relevant code

- `router/src/creator.rs:184` — `round_counter: AtomicU64::new(0)`
- `commonware-restaking/node/src/contributor/handler.rs:228` — `if !signed.insert(round)` guards against signing the same round twice

## Suggested fix

Initialize the round counter from a persistent source rather than always starting at 0. Options:

1. **Read `stateTransitionCount` from the target contract on startup** — ensures the round counter is at least as high as the last successfully executed round
2. **Persist the round counter** (e.g., to a file on the shared PVC) so it survives restarts
3. **Use a monotonic timestamp or block number** as the round identifier instead of a simple counter

PR #287, opened 2026-06-26
fix: seed round counter from wall clock to survive router restarts

bagelface:
## Problem

`ListeningGasKillerCreator`'s round counter reset to `0` on every router restart. Nodes dedup signed rounds in an in-memory `HashSet` and refuse to re-sign a round, so after a restart the router re-proposed rounds the still-running nodes had already signed — stalling aggregation until the nodes were also restarted.

`round` is an opaque, strictly-monotonic consensus sequence number, decoupled from on-chain `stateTransitionCount` and from `transition_index`.

## Fix

Seed `round_counter` from wall-clock milliseconds at startup instead of `0`, so every restart begins above any round a previous run could have dispatched. `fetch_add` keeps it monotonic within a run; at millisecond granularity the seed advances far faster than rounds are produced (each round takes seconds), so the counter stays unique across restarts with no persisted state, RPC, or new config.

This was chosen over the issue's other two options:
- **Read `stateTransitionCount` on startup** — fragile: the counter increments per *dispatched* task (incl. failed/timed-out rounds) and is global, while `stateTransitionCount` is per-target, so a restart could seed *below* the highest round already dispatched and re-collide.
- **Persist to PVC file** — robust but adds I/O, flush-ordering, and a new failure mode; duplicates the epic's future SQLite store.

## Changes

All in `router/src/creator.rs`:
- Add `initial_round_seed()` (ms since epoch; epoch-underflow → `0`).
- Seed `round_counter` from it instead of `0`.
- New unit test `test_initial_round_seed_advances_and_is_nonzero`.

The `Basic` `GasKillerCreator` (hardcoded round `0`) and its tests are unaffected.

## Verification

- `cargo fmt --all`
- `cargo clippy --all-targets --all-features -- -D warnings` (clean)
- `cargo test -p gas-killer-router` — 7/7 creator tests pass, including the new one

### DIFF
--- a/router/src/creator.rs
+++ b/router/src/creator.rs
@@ -13,7 +13,7 @@ use std::collections::HashMap;
 use std::env;
 use std::sync::atomic::{AtomicU64, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
-use std::time::{Duration, Instant};
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
 use tracing::{debug, error, info, warn};
 
@@ -133,6 +133,15 @@ struct EnrichedTask {
     chain_id: u64,
 }
 
+/// Seeds the round counter so it does not restart from a fixed value. Nodes refuse to re-sign a
+/// round, so reusing rounds across restarts would stall aggregation.
+fn initial_round_seed() -> u64 {
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map(|d| d.as_millis() as u64)
+        .unwrap_or(0)
+}
+
 /// Creator for the gas killer usecase that listens for external requests
 pub struct ListeningGasKillerCreator {
     receiver: tokio::sync::Mutex<TaskReceiver>,
@@ -140,7 +149,7 @@ pub struct ListeningGasKillerCreator {
     config: GasKillerConfig,
     validator: Arc<GasKillerValidator>,
     current_task: Mutex<Option<EnrichedTask>>,
-    /// Round counter - incremented for each new task to ensure unique rounds
+    /// Consensus round number, incremented per dispatched task and unique across restarts.
     round_counter: AtomicU64,
     metrics: Option<Arc<MetricsCollector>>,
     /// Shared with the executor to measure P2P round-trip duration.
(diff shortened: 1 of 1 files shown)


BreadrichEngels (approved):
Nice, pragmatic fix for #122 — well-reasoned and correctly scoped. LGTM.

**The core approach is sound.** Seeding `round_counter` from wall-clock ms means each restart starts far above any round a prior run could have reached, so the in-memory node dedup never sees a reused round. The headroom argument holds: the counter only advances by 1 per *dispatched* round and rounds take seconds, so it climbs at ≲1/sec, while the wall-clock seed climbs 1000/sec — any later restart's seed comfortably exceeds the previous run's final counter value. No persisted state, no RPC, no new config. The alternatives analysis (stateTransitionCount is per-target + tracks only successful transitions vs. a global per-dispatch counter; PVC persistence adds I/O and a new failure mode) is accurate, and the roadmap#15 overlap note is right — the future SQLite store persists tasks/keys, not this counter, so it wouldn't close #122 on its own.

**I checked the one thing that could have silently broken: the seed (~1.75e12) now exceeds `u32::MAX`.** If `round` were ever narrowed to `u32` on the wire, in node dedup, or in a contract arg, this would truncate and reintroduce collisions. It doesn't — `round` is `u64` end to end in both this repo and commonware-restaking (`handle_verification(round: u64)`, `message.round`, `HashMap<u64, RoundState>`, `Option<u64>`); the only `u32` on that path is `current_block_number`, which is block height, not round. So the large seed is safe. The `as u64` on `as_millis()` (u128) is likewise fine for ~584M years.

Even the rolling-restart overlap is safe, for what it's worth: with the 360s `terminationGracePeriodSeconds`, a new router can run alongside a draining old one, but the new pod seeds at ~`now_ms` while the old pod's counter is still down near `(now - 360000) + small`, so their round ranges can't meet.

**One residual caveat (non-blocking, and you already point at the robust fix):** wall-clock time isn't monotonic across restarts. A large backward clock step (NTP correction, manual set, VM migration), or significant clock skew between the old and new pod during that 360s drain overlap, could in principle reseed at/below a previously dispatched round and bring back the exact stall this fixes. The ms-vs-≤1/sec headroom makes it very unlikely in practice, and persistence is the real cure — worth a one-line comment in `initial_round_seed` noting the monotonic-clock assumption so the tradeoff is explicit for the next reader.

`test_initial_round_seed_advances_and_is_nonzero` is a reasonable guard (and good that it uses `>=`, so two calls landing in the same millisecond won't flake); the cross-restart uniqueness property itself isn't unit-testable, so that's fine.

CI is `unstable` with no checks reporting yet — confirm `fmt`/`clippy`/`test` are green before merge. Approving.

jesmith0 (dismissed):

PR #289, opened 2026-06-29
feat: support byte array for calldata field in scenario scripts

bagelface:


BreadrichEngels (approved):
Clean, well-tested dev-tooling change — `parse_call_data` now accepts a bracketed decimal byte array in addition to hex, scoped entirely to `scripts/run_scenario.rs` with no production impact. LGTM.

The parsing logic is correct and the edge cases are handled:
- `[]` and a trailing comma both work via `filter(|t| !t.is_empty())` (empty splits dropped), and inner whitespace via `map(str::trim)`.
- Out-of-range (`[256]`) and non-numeric (`[1,foo,3]`) bytes surface as errors through the `parse::<u8>()` + `collect::<Result<_,_>>()` short-circuit, with a helpful `byte "..."` prefix.
- The hex branch keeps the optional-`0x` behaviour and now maps the error to `String` so both branches share one return type.
- Tests cover all of the above (hex w/ and w/o prefix, empty, byte array w/ whitespace+trailing comma, empty array, and the three invalid cases). Good coverage.

I also confirmed these tests actually run: `scripts/Cargo.toml` declares `[[bin]] name = "run_scenario"`, so it's a real cargo target and `cargo test` compiles + executes the `#[test]`s rather than them being dead code in a loose script.

**One trivial nit (non-blocking):** a half-bracketed input like `"[1,2,3"` (or `"1,2,3]"`) fails the `strip_prefix('[') + strip_suffix(']')` pair, so it falls through to the hex branch and reports a confusing hex-decode error rather than a byte-array one. Only matters for a fat-fingered scenario file, and the message still surfaces a failure — fine to leave, or detect a lone leading `[` and give a clearer "malformed byte array" message if you ever touch this again.

CI shows no checks reporting yet (`unstable`) — confirm green before merge. Approving.

PR #290, opened 2026-06-29
feat: use production p2p config for deployed router and node

### LINKED ISSUE #271: Use production p2p config (`lookup::Config::recommended`) for the deployed router/node instead of `::local`
## Problem
The deployed router and node construct their commonware-p2p network with `lookup::Config::local(...)`:
- `router/src/main.rs:168`
- `node/src/main.rs:297`

`Config::local`'s own docstring states it "minimizes peer discovery latency ... useful when running local demos," with an explicit **"It is not recommended to use this configuration in production."** The gas-killer service runs in real clusters (testnet/mainnet on GKE), so the deployed binaries should use the production profile `lookup::Config::recommended(...)`.

Re-homed from BreadchainCoop/commonware-restaking#35 — post-split that repo only governs its local counter example (which correctly uses `Config::local`). The registered-handshakes change discussed there is explicitly **out of scope** here (team decision).

## What differs (`local` → `recommended`, commonware-p2p 0.0.63)
| field | `local` (demo) | `recommended` (prod) |
|---|---|---|
| `allow_private_ips` | **true** | **false** |
| `allowed_connection_rate_per_peer` | `per_second(1)` | `per_minute(1)` |
| `allowed_handshake_rate_per_ip` | `per_second(16)` | 1 per 5s |
| `allowed_handshake_rate_per_subnet` | `per_second(128)` | `per_second(64)` |
| `max_concurrent_handshakes` | 1024 | 512 |
| `ping_frequency` | 5s | 50s |
| `allowed_ping_rate` | `per_second(2)` | `per_minute(2)` |
| `dial_frequency` | 500ms | 1s |
| `query_frequency` | 30s | 60s |

`local` dials/pings aggressively with loose rate limits so a laptop demo converges fast; in a long-running cluster that's wasted churn and weaker abuse protection.

## Proposed change
Switch both binaries from `Config::local(...)` to `Config::recommended(...)`, with two deliberate carve-outs:

1. **Keep `allow_private_ips = true`.** `recommended` sets it `false`, but in-cluster router↔node p2p on GKE resolves to **private pod IPs** — flipping it to `false` would drop all intra-cluster connections. Set `p2p_cfg.allow_private_ips = true` after constructing `recommended` (unless/until the topology uses public addresses). **Verify against the actual deployment networking before merging.**
2. **Leave `attempt_unregistered_handshakes` as-is** (`router/src/main.rs:181`, `node/src/main.rs:310`) — registered-handshakes is out of scope per team decision.

## Files to change
- `router/src/main.rs:168` (constructor), `:181` (preserve handshake override)
- `node/src/main.rs:297` (constructor), `:310` (preserve handshake override)

## Impact / Effort / Risk
- **Impact:** Medium — production-appropriate discovery cadence + rate limits; stops using a profile the library marks unfit for production.
- **Effort:** Low (constructor swap + `allow_private_ips` carve-out).
- **Risk:** Medium — `allow_private_ips`/rate-limit changes can break cluster connectivity if topology isn't accounted for; validate on a testnet soak before mainnet.

## Acceptance Criteria
- Deployed router/node use `lookup::Config::recommended(...)`.
- `allow_private_ips` set appropriately for the cluster topology (router↔node still connect on testnet).
- `attempt_unregistered_handshakes` behavior unchanged.
- No regression in round formation on a testnet soak.

Part of gas-killer/roadmap#15

bagelface:
Closes #271.

## What

Switch the deployed router and node from the demo p2p profile `lookup::Config::local(...)` to the production profile `lookup::Config::recommended(...)` (commonware-p2p 0.0.63), then revert four discovery/rate fields to their `local` values (carve-out #3). `local`'s own docstring marks it unfit for production; the service runs in real GKE clusters.

The **shipped** column is the effective deployed config — not the bare `recommended` column:

| field | local | recommended | shipped (this PR) |
|---|---|---|---|
| `allowed_connection_rate_per_peer` | per_second(1) | per_minute(1) | **per_second(1)** ¹ |
| `allowed_handshake_rate_per_ip` | per_second(16) | 1 per 5s | **per_second(16)** ¹ |
| `allowed_handshake_rate_per_subnet` | per_second(128) | per_second(64) | per_second(64) |
| `max_concurrent_handshakes` | 1024 | 512 | 512 |
| `ping_frequency` | 5s | 50s | 50s |
| `allowed_ping_rate` | per_second(2) | per_minute(2) | per_minute(2) |
| `dial_frequency` | 500ms | 1s | **500ms** ¹ |
| `query_frequency` | 30s | 60s | **30s** ¹ |

¹ reverted to `local` — see carve-out #3.

Net effect: `recommended`'s abuse-resistance (`max_concurrent_handshakes` 512, `allowed_handshake_rate_per_subnet` 64/s, ping cadence 50s / per_minute(2)) **plus** `local`'s discovery aggressiveness for the per-peer / per-ip / dial / query fields.

**No silent off-table adoption.** The only fields that differ between `local` and `recommended` are the eight above plus `allow_private_ips`. Every other `Config` field — `mailbox_size`, `synchrony_bound`, `max_handshake_age`, `handshake_timeout`, `tracked_peer_sets`, `max_message_size` — is identical in both profiles, so the `local`→`recommended` swap changes nothing beyond these tabled deltas.

## Carve-outs

1. **`allow_private_ips = true`** — `recommended` defaults it to `false`, but in-cluster router↔node p2p on GKE resolves to private pod IPs, so leaving it false would drop every intra-cluster connection. Set back to `true` after construction.
2. **`attempt_unregistered_handshakes` unchanged** — registered-handshakes is out of scope per team decision.
3. **Fast peer (re)discovery — `dial_frequency`, `query_frequency`, `allowed_connection_rate_per_peer`, `allowed_handshake_rate_per_ip` reverted to `local` values.** gas-killer is a small, static, allowlisted full mesh: every participant dials every other, so dual-dial reservation collisions are routine and the loser must re-dial fast, and an operator pod restart must rejoin the signing quorum in seconds. `recommended`'s discovery throttling (tuned for large open gossip networks) stretched mesh (re)convergence to ~60–110s — long enough that the round timed out before any node was reachable. This is what failed CI (E2E + Helm both timed out with `currentSum` unchanged) and would, in prod, drop a restarted operator out of the quorum for ~a minute and throttle NAT'd external operators that share an egress IP. Reverting these four restores fast convergence while keeping the abuse-resistance above.

## Files

- `router/src/main.rs` — constructor swap + carve-outs 1–3
- `node/src/main.rs` — constructor swap + carve-outs 1–3 (symmetric with the router so both ends agree on the dial/handshake rates)

## Validation

`cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo check`, and `cargo test` pass for the affected crates. CI is green on the latest commit, including **E2E Test** and **Helm Integration Test** (both regressed on the bare-swap commit and pass after carve-out #3).

The change lives inside `fn main()` (network config construction) with no extractable pure function, so there is no unit-test hook. Per the issue's acceptance criteria, **verify cluster connectivity on a testnet soak before mainnet** — confirm round formation has no regression, and explicitly test the disruption path (`kubectl delete pod` on one operator → it should rejoin the quorum in seconds without stalling round formation).

Part of gas-killer/roadmap#15.

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -296,13 +296,18 @@ fn main() {
         const MAX_MESSAGE_SIZE: usize = 1024 * 1024; // 1 MB
         let my_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), port);
         let my_local_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), port);
-        let mut p2p_cfg = lookup::Config::local(
+        let mut p2p_cfg = lookup::Config::recommended(
             signer.clone(),
             APPLICATION_NAMESPACE,
             my_addr,
             my_local_addr,
             MAX_MESSAGE_SIZE,
         );
+
+        // recommended() sets this false, but in-cluster router<->node p2p on GKE resolves to private
+        // pod IPs; leaving it false would drop every intra-cluster connection. Keep it true until the
+        // topology uses public addresses.
+        p2p_cfg.allow_private_ips = true;
 
         // Must stay true for K8s deployments (DNAT/SNAT means source IPs at the listener are
         // always pod IPs, never the registered ClusterIP addresses) and for mixed-network topologies
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -13,6 +13,7 @@ use commonware_avs_node::contributor::{AggregationInput, Contribute, Contributor
 use commonware_p2p::Manager;
 use commonware_p2p::authenticated::lookup::{self, Network};
 use commonware_runtime::{Metrics, Runner, Spawner, tokio};
+use commonware_utils::NZU32;
 use commonware_utils::set::OrderedAssociated;
 use eigen_logging::log_level::LogLevel;
 use gas_killer_common::{
@@ -315,6 +316,18 @@ fn main() {
         // case; authentication relies entirely on the cryptographic handshake (peer public keys
         // checked against the registered operator set), which is secure for both topologies.
         p2p_cfg.attempt_unregistered_handshakes = true;
+
+        // recommended() throttles peer discovery for large open gossip networks where aggressive
+        // dialing is abusive. gas-killer instead runs a small, static, allowlisted operator set in a
+        // full mesh: every participant dials every other, so both ends frequently dial at once and
+        // one connection loses the reservation race. The loser must re-dial quickly, and an operator
+        // that restarts must rejoin the signing quorum in seconds rather than ~a minute. Restore fast
+        // (re)discovery while keeping recommended's abuse-resistance (concurrent-handshake cap, subnet
+        // rate limit, ping cadence).
+        p2p_cfg.dial_frequency = Duration::from_millis(500);
+        p2p_cfg.query_frequency = Duration::from_secs(30);
+        p2p_cfg.allowed_connection_rate_per_peer = Quota::per_second(NZU32!(1));
+        p2p_cfg.allowed_handshake_rate_per_ip = Quota::per_second(NZU32!(16));
 
         let (mut network, mut oracle) = Network::new(context.with_label("network"), p2p_cfg);
 
--- a/router/src/main.rs
+++ b/router/src/main.rs
@@ -15,6 +15,7 @@ use commonware_runtime::{
     Metrics, Runner, Spawner,
     tokio::{self},
 };
+use commonware_utils::NZU32;
 use commonware_utils::set::OrderedAssociated;
 use eigen_logging::log_level::LogLevel;
 use gas_killer_common::{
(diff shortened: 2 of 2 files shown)


BreadrichEngels (approved):
Good move off the demo `local` profile, and the reasoning for the hybrid (recommended's abuse caps + a static full-mesh's need for fast re-dial) is sound. The code is correct and — importantly — the router and node carve-outs are symmetric, so both ends agree on the rate limits (mismatched dial/handshake rates between the two would have one side throttling the other). Approving, with one description-accuracy request and the usual soak caveat.

**Please reconcile the PR description with the diff.** As written, the "What" table presents the full `recommended` column as what's being adopted, and "Carve-outs" lists only two (`allow_private_ips`, `attempt_unregistered_handshakes`). But the second commit adds a *third* carve-out block that reverts four of the eight tabled fields back to their `local` values:

| field | recommended | this PR actually ships |
|---|---|---|
| `dial_frequency` | 1s | **500ms** (local) |
| `query_frequency` | 60s | **30s** (local) |
| `allowed_connection_rate_per_peer` | per_minute(1) | **per_second(1)** (local) |
| `allowed_handshake_rate_per_ip` | 1 per 5s | **per_second(16)** (local) |

So the effective config is "recommended's `max_concurrent_handshakes` (512), `allowed_handshake_rate_per_subnet` (64/s), and ping cadence (50s/per_minute(2)) — but `local`'s discovery aggressiveness." That's a reasonable hybrid and the inline comment justifies it well; the issue is just that someone reading the PR/issue#271 later would conclude the deployed per-peer and per-ip limits are the tight recommended ones when they aren't. Add the four reverts as carve-out #3 (or annotate the table) so the deployed limits are unambiguous.

**Notes / for the soak:**
- `allow_private_ips = true` is necessary for GKE pod IPs and consistent with the existing "security rests on the crypto handshake allowlist, not IP filtering" model (same rationale as `attempt_unregistered_handshakes`). Fine.
- `recommended()` likely sets more than the eight fields in the table (timeouts, buffers, etc.) that you're now silently adopting. Worth a quick confirm that none of those off-table changes hurt the small-mesh/quorum-formation path — exactly what the testnet soak should catch.
- As the PR says, this is constructed inline in `fn main()` with no pure-function hook, so there's no unit test — the **testnet soak before mainnet (confirm round formation has no regression)** is the real gate here. Don't skip it; p2p rate/discovery changes are the kind of thing that looks fine at compile and bites under real cluster timing.

CI shows no checks reporting yet — confirm `fmt`/`clippy`/`check`/`test` green before merge. Approving on the code; just sync the description and run the soak.

bagelface:
Confirmed on testnet deploy https://sepolia.etherscan.io/tx/0x7953db89cf3bd17c4092a3de796b0b589f845912421867bfc7730560fa6ea29e

Issue #193, opened 2026-05-05
Async task submission returns task IDs instead of fire-and-forget


bagelface:
## Parent Epic
gas-killer/roadmap#15

## Background

The current `POST /trigger` returns `{ success, message }` immediately after enqueueing. Users have no way to track what happened to their request after acknowledgement. Switching to an async model where submission returns a task ID enables proper status tracking (TASK-2) and the new payload delivery model (TASK-3).

## Acceptance Criteria

- [ ] `POST /tasks` replaces `POST /trigger` as the primary submission endpoint
- [ ] Successfully queued tasks return `202 Accepted` with body: `{ "task_id": "<uuid-v4>", "status": "queued" }`
- [ ] Task IDs are UUID v4 and globally unique
- [ ] All existing request validation (addresses, calldata size, onchain checks) still runs before accepting
- [ ] Validation errors remain synchronous: `400` / `422` / `503` as today
- [ ] `POST /trigger` is kept as a deprecated alias returning the same shape, to ease migration

## Metadata

| Field | Value |
|-------|-------|
| Epic | Task Lifecycle Redesign |
| Story ID | TASK-1 |
| Complexity | Medium |
| Priority | P0 — must-ship |

bagelface:
**Note from #199 / #292 (structured error codes):** this story owns the convergence of the **success** response shape.

As of #199, error responses use the envelope `{ "error": { "code", "message" } }`, but the success path is still `{ "success": true, "message": "..." }`. That's an intentional transient asymmetry — TASK-1 replaces the success body here with `202 { "task_id": "<uuid>", "status": "queued" }`, at which point both success and error shapes are deliberate and stable.

Two follow-ups when implementing this:
- Drop the legacy `{ success, message }` success body (the deprecated `POST /trigger` alias should return the same `202 { task_id, status }` shape).
- The two-shape window between #199 and this story must be documented accurately in #201 (API reference) and #202 (quickstart) so beta integrators branch on HTTP status first, then parse the matching body.

Issue #191, opened 2026-05-05
Global queue depth cap with backpressure


bagelface:
## Parent Epic
gas-killer/roadmap#15

## Background

The current task queue is an unbounded in-memory `Vec`. A burst of valid requests can queue indefinitely, causing unbounded memory growth and tasks being processed hours after submission (when their block references are long stale).

## Acceptance Criteria

- [ ] Configurable max queue depth via `MAX_QUEUE_DEPTH` env var (default: 100)
- [ ] Requests submitted when the queue is at capacity return `503 Service Unavailable` with error code `QUEUE_FULL` and a `Retry-After` estimate
- [ ] The queue depth check and enqueue are atomic (no TOCTOU race)
- [ ] Current queue depth is exposed as a Prometheus gauge: `gas_killer_queue_depth`

## Metadata

| Field | Value |
|-------|-------|
| Epic | Rate Limiting |
| Story ID | RATE-2 |
| Complexity | Low |
| Priority | P0 — must-ship |

Issue #190, opened 2026-05-05
Per-API-key rate limiting on /tasks


bagelface:
## Parent Epic
gas-killer/roadmap#15

## Background

No HTTP-level rate limiting exists today. A single misbehaving client can saturate the ingress queue and block all other users. The `governor` crate is already a dependency for P2P rate limiting and can be reused here.

## Acceptance Criteria

- [ ] Configurable global default rate limit per key: requests per minute via `RATE_LIMIT_RPM` env var (default: 60)
- [ ] Requests exceeding the limit return `429 Too Many Requests` with a `Retry-After` header (seconds until next allowed request)
- [ ] Rate limit state is keyed by `key_id`, not IP address
- [ ] Individual keys can be issued with a custom RPM override (set at creation time via `POST /admin/keys`)
- [ ] Rate limit counters are in-memory (sliding window); a restart resets them — acceptable for v1
- [ ] Rejections increment a counter metric (see OPS-1)

## Notes

> **Implementation note:** Use a keyed `governor::RateLimiter` with a `DashMap` backing store. The existing P2P use of `governor::Quota::per_second` is a direct reference point.

## Metadata

| Field | Value |
|-------|-------|
| Epic | Rate Limiting |
| Story ID | RATE-1 |
| Complexity | Low |
| Priority | P0 — must-ship |

Issue #188, opened 2026-05-05
Token-based API key authentication with revocation


bagelface:
## Parent Epic
gas-killer/roadmap#15

## Background

Currently auth is a single shared Bearer token set at deploy time via `INGRESS_PASSWORD`. There is no per-client identity and no way to revoke access without redeploying. Replace with an API key system that assigns an opaque key per client and allows revocation without a service restart.

## Acceptance Criteria

- [ ] `POST /admin/keys` (admin-only) creates an API key and returns `{ id, key, label, created_at }` — the key value is shown exactly once
- [ ] `DELETE /admin/keys/{id}` revokes the key immediately; subsequent requests using it return 401
- [ ] `GET /admin/keys` lists active key IDs and metadata (`created_at`, `label`, `last_used`) — never the key value
- [ ] All `/tasks` requests require a valid, unrevoked key via `Authorization: Bearer <key>`
- [ ] Admin endpoints are protected by a separate credential (`ADMIN_KEY` env var)
- [ ] Key store survives router restarts (see #RELY-1 / RELY-1)
- [ ] Constant-time key comparison is preserved to prevent timing attacks

## Notes

> The admin credential can remain a single env-var secret since it is for operator use only. Only user-facing keys need full CRUD. Key values should be generated with a cryptographically secure RNG and formatted as `gk_<32-byte hex>` for easy identification in logs.

## Metadata

| Field | Value |
|-------|-------|
| Epic | Authentication & Authorization |
| Story ID | AUTH-1 |
| Complexity | Medium |
| Priority | P0 — must-ship |

bagelface:
Depends on #293 (scaffold SQLite persistence layer). The `keys` table and `/admin/keys` CRUD here build on the `store` module + migration runner introduced in #293, rather than introducing the DB layer in this PR. The raw key is shown once on create; only a hash is persisted.

PR #292, opened 2026-06-29
feat: structured machine-readable error codes (OPS-2)

### LINKED ISSUE #199: Structured, machine-readable error codes
## Parent Epic
gas-killer/roadmap#15

## Background

Current error responses are plain strings (`{ "success": false, "message": "target_address must be non-zero" }`). Beta clients building integrations need to distinguish error types programmatically without parsing free text.

## Acceptance Criteria

- [ ] All error responses follow a consistent envelope: `{ "error": { "code": "...", "message": "..." } }`
- [ ] Error codes are a documented, stable, uppercase snake-case enum
- [ ] HTTP status codes are unchanged from current behavior
- [ ] Error codes cover at minimum: `INVALID_ADDRESS`, `STALE_BLOCK`, `TRANSITION_MISMATCH`, `CALLDATA_TOO_LARGE`, `RATE_LIMITED`, `QUEUE_FULL`, `UNAUTHORIZED`, `RPC_UNAVAILABLE`

## Error Shape

```json
{
  "error": {
    "code": "TRANSITION_MISMATCH",
    "message": "expected transition_index 42, contract reports 43"
  }
}
```

## Metadata

| Field | Value |
|-------|-------|
| Epic | Observability & Operations |
| Story ID | OPS-2 |
| Complexity | Low |
| Priority | P1 — strong recommendation |

bagelface:
Closes #199.

Replaces the ad-hoc `{ "success": false, "message": "..." }` error responses on the ingress with a consistent, machine-readable envelope so beta integrators can branch on error type without parsing free text.

## What changed

**New `router/src/error.rs`:**
- `ErrorCode` — a stable, documented enum serialized as `SCREAMING_SNAKE_CASE`. Covers all eight codes the issue requires (`INVALID_ADDRESS`, `STALE_BLOCK`, `TRANSITION_MISMATCH`, `CALLDATA_TOO_LARGE`, `RATE_LIMITED`, `QUEUE_FULL`, `UNAUTHORIZED`, `RPC_UNAVAILABLE`) plus `CONTRACT_NOT_FOUND`, `INVALID_REQUEST`, `INTERNAL` for the existing handler paths. `RATE_LIMITED`/`QUEUE_FULL` are defined ahead of their emitters (#190/#191) so the enum stays the single source of truth.
- `ApiError` (status + code + message) implements `IntoResponse`, serializing to `{ "error": { "code", "message" } }`.
- `ApiJson<T>` — a drop-in `Json<T>` extractor that routes body-parse / schema-validation rejections through the same envelope while preserving axum's status code (400 malformed, 422 shape mismatch), so *every* error response — handler-level and extractor-level — shares the contract.

**`router/src/ingress.rs`:**
- `From<ValidationError>` / `From<OnchainValidationError>` for `ApiError` centralize the status+code mapping. RPC failures keep their sanitized `"Service temporarily unavailable"` message — internal RPC detail is not leaked to clients.
- `trigger_task_handler` returns `Result<_, ApiError>` and uses `ApiJson`. All existing `warn!` logging and metric increments are preserved; only the error body shape changed.

## Acceptance criteria
- [x] Consistent `{ "error": { "code", "message" } }` envelope across all error responses (handler + extractor)
- [x] Stable, documented, uppercase snake-case code enum
- [x] HTTP status codes unchanged from prior behaviour
- [x] All eight required codes present

## Notes
- The **success** response shape is intentionally left as-is; #193 redefines it (`202 { task_id, status }`).
- This is the Phase-1 foundation the rest of the v1-beta lifecycle builds on: #188 uses `ApiError::unauthorized()`, #191 emits `QUEUE_FULL`, #190 emits `RATE_LIMITED`, and #193/#194's new endpoints get the envelope for free via `ApiJson`.

## Verification
- `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --all-targets --all-features` all pass across the workspace.
- New unit tests cover code serialization, the envelope shape, the `From` mappings (incl. RPC-detail sanitization and future-vs-stale block distinction); every error-path HTTP test now asserts on `error.code`.
- No new runtime config, so no `example.env` / Helm changes.

### DIFF
--- a/router/src/error.rs
+++ b/router/src/error.rs
@@ -47,6 +47,10 @@ pub enum ErrorCode {
     ContractNotFound,
     /// The request body is malformed or fails field-level validation.
     InvalidRequest,
+    /// The requested path does not exist.
+    NotFound,
+    /// The HTTP method is not supported for the requested path.
+    MethodNotAllowed,
     /// An unexpected server-side error occurred.
     Internal,
 }
@@ -169,6 +173,8 @@ mod tests {
             (ErrorCode::RpcUnavailable, "RPC_UNAVAILABLE"),
             (ErrorCode::ContractNotFound, "CONTRACT_NOT_FOUND"),
             (ErrorCode::InvalidRequest, "INVALID_REQUEST"),
+            (ErrorCode::NotFound, "NOT_FOUND"),
+            (ErrorCode::MethodNotAllowed, "METHOD_NOT_ALLOWED"),
             (ErrorCode::Internal, "INTERNAL"),
         ];
         for (code, wire) in cases {
--- a/router/src/error.rs
+++ b/router/src/error.rs
@@ -1,0 +1,211 @@
+//! Machine-readable error envelope for the ingress HTTP API.
+//!
+//! Every error response the API returns has the shape:
+//!
+//! ```json
+//! { "error": { "code": "TRANSITION_MISMATCH", "message": "..." } }
+//! ```
+//!
+//! The `code` is a stable, uppercase snake-case identifier that integrators match on
+//! programmatically; the `message` is a human-readable explanation that may change at
+//! any time. Status codes are carried alongside the code internally and are unchanged
+//! from the per-endpoint behaviour they replace.
+
+use axum::extract::rejection::JsonRejection;
+use axum::extract::{FromRequest, Request};
+use axum::http::StatusCode;
+use axum::response::{IntoResponse, Response};
+use axum::{Json, async_trait};
+use serde::{Deserialize, Serialize};
+
+/// Stable, machine-readable error code returned in every API error response.
+///
+/// These identifiers are a public API contract: integrators branch on them, so a variant
+/// must never be renamed or repurposed once shipped. Add a new variant rather than changing
+/// the meaning of an existing one. The wire form is uppercase snake-case (e.g.
+/// `TRANSITION_MISMATCH`).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum ErrorCode {
+    /// A required address field is zero or otherwise not a usable address.
+    InvalidAddress,
+    /// `block_height` is older than the accepted staleness window.
+    StaleBlock,
+    /// The provided `transition_index` does not match the contract's current state.
+    TransitionMismatch,
+    /// `call_data` exceeds the maximum accepted size.
+    CalldataTooLarge,
+    /// The client exceeded its allotted request rate.
+    RateLimited,
+    /// The ingress queue is at capacity and cannot accept more work right now.
+    QueueFull,
+    /// The request is missing valid authentication credentials.
+    Unauthorized,
+    /// An upstream RPC endpoint failed or is unreachable.
+    RpcUnavailable,
+    /// No contract is deployed at the requested target address on any supported chain.
+    ContractNotFound,
+    /// The request body is malformed or fails field-level validation.
+    InvalidRequest,
+    /// An unexpected server-side error occurred.
+    Internal,
+}
+
+/// The `error` object nested inside [`ApiErrorEnvelope`].
+#[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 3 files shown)


BreadrichEngels (approved):
Excellent, well-architected change — the envelope is cleanly factored, the `From` mappings centralize status+code in one place, and the test coverage is genuinely thorough (per-path `error.code` assertions, the `From` mapping unit tests, and RPC sanitization all locked in). Meets every acceptance criterion in #199. LGTM.

Things I specifically checked that hold up:

- **Status codes are preserved 1:1 with the old handler.** 401 (auth), 503 (queue full), 400 (validation + most onchain), 503 (RpcError), 500 (channel closed), 200 (success) all map through unchanged. `BlockHeightInFuture` correctly stays 400/`INVALID_REQUEST` (not `STALE_BLOCK`), matching the prior `else → 400` branch.
- **RPC detail is sanitized.** `OnchainValidationError::RpcError` → fixed `"Service temporarily unavailable"` rather than `e.to_string()`, with a test asserting an internal IP doesn't leak. Good — that's the kind of thing that silently regresses, and you pinned it.
- **The 404/405 framework wrapping is handled carefully.** Replacing the body in place (rather than rebuilding the response) preserves the `Allow` header a spec-compliant 405 must carry — and you've got a test asserting `Allow` still lists POST after the rewrite, which is exactly the regression I'd worry about with response middleware. Resetting `CONTENT_LENGTH` alongside the new `CONTENT_TYPE` is the right detail. The serialize-failure fallback (return the original response instead of panicking) is a nice touch.
- **Extractor errors now share the contract.** `ApiJson` routing `JsonRejection` through the same envelope while preserving the rejection's status (400 malformed / 422 shape / 415 media-type) closes the old gap where malformed-JSON responses were plain text. And since the body was always parsed by the extractor before the handler ran, there's no change to when load-shedding happens — the capacity check still precedes the expensive `validate_onchain` RPC, so #273's shed-before-RPC property is intact.

**Minor / non-blocking:**
- The `wrap_framework_error` heuristic keys on "404/405 with no `Content-Type` ⇒ framework-generated." That's correct today (no handler emits a bodyless 404/405) and you document it in the fn comment, but it's an implicit contract — if a future handler ever returns a bare `StatusCode::NOT_FOUND`/`METHOD_NOT_ALLOWED`, it'll get silently re-wrapped into the envelope. That's arguably desirable, just worth keeping in mind as the API grows.
- Transient shape asymmetry: success stays `{ success, message }` while errors are `{ error: { code, message } }`, so a beta integrator sees two shapes until #193 redefines success. Intentional and documented — fine, just flagging for the integration docs.
- `ErrorCode::RateLimited` is defined but not yet emitted (its emitter is #190); harmless and the "enum is the single source of truth" rationale is reasonable.

CI shows no checks reporting on the head commit yet — confirm `fmt`/`clippy -D warnings`/`test` are green before merge (you note they pass locally). Approving.

bagelface:
Thanks for the review! Following up on the two flags:

**1. `wrap_framework_error` heuristic ("404/405 + no `Content-Type` ⇒ framework-generated").** Good call on the implicit contract — I've made it explicit and CI-guarded in `94f3cb4`. The new test `test_bare_status_handler_is_wrapped_but_envelope_handler_is_preserved` pins both directions:
- a handler returning a bare `StatusCode::NOT_FOUND` (no body/Content-Type) → rewritten into the generic envelope, and
- a handler returning an `ApiError` envelope (Content-Type `application/json`) → passed through with its specific status/code/message intact (e.g. `"widget 7 not found"` is *not* clobbered to `"Not found"`).

So the re-wrap is the intended, now-tested behavior: a future handler physically can't leak a non-enveloped bare 404/405, while any handler that builds its own envelope (e.g. #194's `GET /tasks/{id}` task-not-found) is untouched.

**2. Success vs error shape asymmetry (`{ success, message }` vs `{ error: { code, message } }`).** Leaving this as-is by design: #193 (TASK-1) redefines the success body to `202 { task_id, status }`, so harmonizing it now would just be thrown away next story. I'll make sure the two-shape window is documented accurately in #201 (API reference) / #202 (quickstart), and I'll note on #193 that it converges the success shape.

Issue #192, opened 2026-05-05
Persistent task store (survive restarts)


bagelface:
## Parent Epic
gas-killer/roadmap#15

## Background

The current in-memory task queue is silently lost on any router restart. For a beta where users are polling for task status, a pod restart must not cause submitted tasks to disappear.

## Acceptance Criteria

- [ ] Task state (queue position, status, payload, metadata) is written to a persistent store before the HTTP response is sent
- [ ] API keys (AUTH-1) are stored in the same persistent store
- [ ] On startup, tasks in `queued` or `processing` state are re-queued automatically
- [ ] Persistent store path is configurable via `DATA_DIR` env var; defaults to `/data`
- [ ] Store uses SQLite for v1 beta (single-replica router; can be swapped for an external DB later)
- [ ] Helm chart mounts a PVC at `/data` for the router pod

## Notes

> SQLite is the lowest-complexity option for a single-replica router. If the service ever scales to multiple replicas, this can be swapped for an external Postgres or Redis instance with no API surface change.

## Metadata

| Field | Value |
|-------|-------|
| Epic | Reliability & Persistence |
| Story ID | RELY-1 |
| Complexity | Medium |
| Priority | P0 — must-ship |

bagelface:
Depends on #293 (scaffold SQLite persistence layer), which introduces the `sqlx` `SqlitePool`, migration runner, `DATA_DIR` config, and the Helm PVC mount at `/data`. This issue then adds the `tasks` table and the re-queue-of-in-flight-tasks-on-startup behaviour on top. The PVC-mount acceptance criterion is satisfied by #293.

bagelface:
Follow-on note for the startup re-queue path: `incomplete_tasks()` (added in the task model PR #300) loads all `queued`/`processing` tasks in a single unbounded `SELECT ... ORDER BY created_at`. Fine at beta volume, but when wiring automatic re-queue on startup, consider paging the backlog (LIMIT/OFFSET or keyset) so a large in-flight set doesn't load in one shot.
Tracked in issue: https://github.com/gas-killer/service/issues/301

bagelface:
All acceptance criteria are satisfied on `task-lifecycle-refactor`:

- Write-before-ack: `submit_task_handler` persists the task via `store.create_task` and gets a task id back before sending the `202` response.
- API keys share the same SQLite store as tasks (`store/api_keys.rs`, `store/tasks.rs`).
[...]

PR #294, opened 2026-06-30
feat: scaffold sqlite persistence layer for the router

### LINKED ISSUE #293: Scaffold SQLite persistence layer (store module + migration runner)
## Parent Epic
gas-killer/roadmap#15

## Background

The v1-beta epic puts two persistent stores in a single SQLite database: API keys (AUTH-1, #188) and tasks (RELY-1, #192). The router has **no persistence layer today** — there is no DB crate anywhere in the workspace.

Introducing `sqlx`, the connection pool, the migration runner, and `DATA_DIR` config inside the auth PR would conflate foundation with feature. Scaffold the shared persistence layer first so #188 and #192 each add only their own table on top of it.

## Acceptance Criteria

- [ ] `sqlx` added to `[workspace.dependencies]` (`sqlite`, `migrate`, `runtime-tokio` features) and referenced from the `router` crate via `{ workspace = true }`
- [ ] New `router/src/store/` module that builds a `SqlitePool` at startup
- [ ] Connection pragmas set via `SqliteConnectOptions`: `journal_mode=WAL`, `foreign_keys=ON`, a non-zero `busy_timeout`
- [ ] `migrations/` directory with the `sqlx::migrate!` runner invoked on startup; applies pending migrations idempotently; **startup aborts loudly if a migration fails**
- [ ] Database path resolved from the `DATA_DIR` env var (default `/data`) → `<DATA_DIR>/router.db`; tests use an in-memory or tempfile database
- [ ] Pool handle reachable from `IngressState`/orchestrator (no domain queries yet)
- [ ] Helm chart mounts a PVC at `/data` for the router pod, and `DATA_DIR` is wired through chart values (satisfies the PVC-mount criterion of #192)
- [ ] Tracing on store init (resolved path, count of migrations applied)
- [ ] Test: a fresh database migrates cleanly, re-running migrations is a no-op, and a trivial `SELECT 1` round-trips through the pool

## Out of Scope (owned by consumers)

- `keys` table + `/admin/keys` CRUD → #188 (AUTH-1)
- `tasks` table + re-queue-of-in-flight-tasks on startup → #192 (RELY-1)

## Notes

> Use the **runtime-checked** `sqlx::query` / `sqlx::query_as` API (with `#[derive(sqlx::FromRow)]`), **not** the compile-time `query!` / `query_as!` macros. The compile-time macros require a live `DATABASE_URL` at build time or a checked-in `.sqlx/` offline cache — friction we don't want layered on top of the private-git-dep (`GH_PAT`) CI build. The runtime API gives the same pooling and migrations with zero build-time database requirement.
>
> SQLite is bundled via `libsqlite3-sys` (no system `libsqlite3` needed at runtime), but it compiles C from a vendored amalgamation — confirm the `router/Dockerfile` builder stage has a C toolchain.

## Metadata

| Field | Value |
|-------|-------|
| Story ID | RELY-0 |
| Complexity | Small–Medium |
| Priority | P0 — blocks #188 and #192 |

bagelface:
Closes #293.

Scaffolds the shared SQLite persistence layer the router lacked entirely, so the API-key (#188) and task-store (#192) features each add only their own table on top of it instead of introducing the DB layer inside a feature PR.

## What's here

**Persistence layer** — `router/src/store/mod.rs`
- `SqliteStore` opens a `SqlitePool` via `SqliteConnectOptions` with `journal_mode=WAL`, `foreign_keys=ON`, and a 5s `busy_timeout`.
- `connect()` resolves `DATA_DIR` (default `/data`) → `<DATA_DIR>/router.db`, creates the parent dir, then runs the embedded `sqlx::migrate!("./migrations")` runner. Migrations are compiled into the binary, so nothing migration-related ships in the runtime image.
- Uses the **runtime-checked** `sqlx::query`/`query_scalar` API (not the compile-time `query!` macros), so CI needs no `DATABASE_URL` or `.sqlx` offline cache.

**Wiring**
- `IngressState` gains `store: Option<SqliteStore>` + a `with_store()` builder (existing test constructors default to `None`).
- `create_listening_creator_with_server` opens the store **before** serving traffic; a migration/open failure propagates to `.expect("Failed to build orchestrator")` and aborts startup loudly rather than serving against an unmigrated store.

**Deployment**
- New `router-data-pvc.yaml` + `router.persistence` values; PVC mounted RW at `/data`, `DATA_DIR` wired through. `fsGroup: 1000` so the non-root `appuser` can write — applied in both PVC and `emptyDir` (disabled) modes.
- `docker-compose.yml` overrides `DATA_DIR=/app/data` (an `appuser`-writable path) so e2e/local don't hit the unwritable `/data`.
- Documented in `example.env`; runtime DB files ignored in `.gitignore`.

**CI** — `helm-integration-tests.yml` gains a *Verify router persistence* step asserting `/data/router.db` exists in-cluster, so a persistence regression surfaces as a named failure instead of a generic readiness timeout.

## Testing
- `cargo fmt` · `clippy --all-targets --all-features -D warnings` · `cargo test` (**81 router tests**, +3 new store tests covering migrate + `SELECT 1`, WAL assertion + idempotent reopen, and the default-path resolution).
- `helm lint` + `helm template` verified in both `persistence.enabled` true/false modes.

No Dockerfile change needed: the builder already links binaries (C toolchain present for the bundled `libsqlite3-sys` compile) and bundled SQLite is statically linked, so the runtime stage needs no system lib.

## Out of scope (owned by consumers)
- `keys` table + `/admin/keys` CRUD → #188 (AUTH-1)
- `tasks` table + re-queue-on-startup → #192 (RELY-1)

## Related
- gas-killer/infra#4 — tracks the zonal-PD/regional-cluster zone-failover gap for this volume (no infra change required to ship).

### DIFF
--- a/.github/workflows/helm-integration-tests.yml
+++ b/.github/workflows/helm-integration-tests.yml
@@ -316,6 +316,20 @@ jobs:
 
           echo "All pods are ready!"
 
+      - name: Verify router persistence
+        run: |
+          # The router opens its SQLite store under DATA_DIR (/data) at startup and aborts if
+          # the volume is missing or not writable by the non-root user. Assert the database
+          # file exists so a persistence regression surfaces here instead of as a generic
+          # readiness timeout or a downstream task failure.
+          ROUTER_POD=$(kubectl get pods -l app.kubernetes.io/component=router -o jsonpath='{.items[0].metadata.name}')
+          echo "Router pod: $ROUTER_POD"
+          echo "=== Contents of /data ==="
+          kubectl exec "$ROUTER_POD" -- ls -l /data
+          echo "=== Asserting /data/router.db exists ==="
+          kubectl exec "$ROUTER_POD" -- test -f /data/router.db
+          echo "router.db is present and the persistent volume is writable"
+
       - name: Setup port forwarding for testing
         run: |
           # Find actual service names (they include the chart name)
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,9 @@ terraform.tfstate
 terraform.tfstate.*
 *.tfstate
 *.tfstate.*
+
+# Router SQLite store — created at runtime under DATA_DIR; never committed.
+# Matches the database file and its WAL/SHM sidecars in any directory.
+router.db
+router.db-*
+/data/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2629,6 +2629,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "concurrent-queue"
+version = "2.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4ca0197aee26d1ae37445ee532fefce43251d24cc7c166799f4d46817f1d3973"
+dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
 name = "const-hex"
 version = "1.18.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3561,6 +3570,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "event-listener"
+version = "5.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e13b66accf52311f30a0db42147dadea9850cb48cd070028831ae5f5d4b856ab"
+dependencies = [
+ "concurrent-queue",
+ "parking",
+ "pin-project-lite",
+]
+
+[[package]]
 name = "eyre"
 version = "0.6.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3675,7 +3695,7 @@ checksum = "da0e4dd2a88388a1f4ccc7c9ce104604dab68d9f408dc34cd45823d5a9069095"
 dependencies = [
  "futures-core",
  "futures-sink",
- "spin",
+ "spin 0.9.8",
 ]
 
 [[package]]
@@ -5938,6 +5958,12 @@ dependencies = [
  "quote",
  "syn 2.0.117",
 ]
+
+[[package]]
+name = "parking"
+version = "2.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f38d5652c16fde515bb1ecef450ab0f6a219d619a7274976324d5e377f7dceba"
 
 [[package]]
 name = "parking_lot"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1342,6 +1342,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "atoi"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f28d99ec8bfea296261ca1af174f24225171fea9664ba9003cbebee704810528"
+dependencies = [
+ "num-traits",
+]
+
+[[package]]
 name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2657,6 +2666,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-queue"
+version = "0.3.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0f58bbc28f91df819d0aa2a2c00cd19754769c2fad90579b3592b1c9ba7a3115"
+dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
 name = "crossbeam-utils"
 version = "0.8.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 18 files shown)


BreadrichEngels (approved):
Clean, well-tested scaffold — the store module is nicely isolated, migrations are embedded so nothing ships in the runtime image, the optional `store` keeps every existing test path untouched, and the CI `test -f /data/router.db` assertion is a good way to turn a persistence regression into a named failure. The store tests are thoughtful too (the single-connection cap for the in-memory DB is exactly the sqlx gotcha you have to handle, and the WAL-pragma + idempotent-reopen file test covers the real path). LGTM on the code.

One operational gap worth addressing, since it directly undercuts what this PR is *for* (surviving restarts/upgrades):

**🚩 RWO PVC + default rolling-update strategy can deadlock the rollout.** I checked the full `router-deployment.yaml`: `replicas: 1` and no `strategy:` field, so it's RollingUpdate (for one replica that's maxSurge 1 / maxUnavailable 0 — surge-first). The new `router-data` PVC is `ReadWriteOnce`. On an upgrade, the surge pod is created before the old one terminates and competes for the same RWO volume:
- same node as the old pod → RWO permits it, fine;
- **different node → Multi-Attach error**, the surge pod is stuck `ContainerCreating`, never goes Ready, so the old pod is never torn down and never releases the volume → the rollout stalls until manual intervention.

The fix is to set the router Deployment to `strategy: { type: Recreate }` (at minimum when `persistence.enabled`). Tradeoff: Recreate means the old pod is gone before the new one starts, so an upgrade incurs brief downtime (≈ the 360s grace + the init-container waits) — but for a single-instance router that's already a single point, bounded downtime is far better than a wedged rollout, and it's the standard pattern for an RWO-backed singleton. Worth doing in this PR since you're introducing the volume here.

A couple of things I confirmed are *fine* (not asks):
- **RWO + WAL is the right pairing.** WAL needs a real filesystem with working shared-memory locking; a block-backed RWO PVC (ext4/xfs) provides that. If anyone later reaches for RWX/NFS to dodge the Multi-Attach issue above, WAL's `-shm` would break — so keep it RWO + Recreate, not RWX.
- `fsGroup: 1000` matching `appuser`, migrate-before-serve with a loud `.expect` abort, and the `DATA_DIR=/app/data` compose override (since `/data` isn't writable locally) all check out.

**Minor:**
- No test covers the `DATA_DIR` override branch (only the default constant) — reasonable, since mutating process env races other tests; the logic is trivial. Fine as-is.
- `fsGroup`/uid `1000` is hardcoded to the current image user — documented, just a latent coupling if the base image's user ever changes.
- Good that gas-killer/infra#4 already tracks the zonal-PD zone-failover gap; no infra change needed to ship this.

CI isn't reporting on the head commit yet — confirm `fmt`/`clippy -D warnings`/`test` plus `helm lint`/`template` are green before merge. This also pulls in `libsqlite3-sys` (C-compiled, bundled/static), so the image build is the thing to watch. Approving on the code; please add the `Recreate` strategy before this is relied on for prod upgrades.

BreadrichEngels (approved):
Re-review of the three commits since my last pass — my blocker is resolved correctly and both minors are handled, plus a nice observability addition. 

**✅ `Recreate` strategy — the blocker is fixed exactly right.** The router Deployment now sets `strategy: { type: Recreate }` gated on `persistence.enabled`, with a comment that captures the RWO/Multi-Attach rationale. Gating it on `persistence.enabled` is the right call — the `emptyDir` (persistence-disabled) path keeps RollingUpdate, which is fine since there's no shared volume to contend for. This closes the rollout-deadlock hazard.

**✅ DATA_DIR override is now tested.** Splitting `resolve_db_path()` into a pure `db_path_from_dir(Option<String>)` and testing both the default and override branches (without mutating process env) is exactly the clean way to cover that — addresses my minor.

**➕ Bonus: `gas_killer_db_up` liveness metric + dashboard panel.** Nice add — seeding the gauge to 1 after `connect()` and re-checking every 15s in a background task, surfaced as a UP/DOWN stat panel, turns a silent volume loss into a visible signal. The `store.clone()` (Arc pool) + `Arc::clone(&metrics)` into the detached task is correct, and the gauge test is a good guard.

**One nuance on that metric (non-blocking):** `health_check()` runs `SELECT 1`, which is a constant expression that doesn't touch the database file — so it reliably detects a dead/closed pool or a detached volume (connection acquisition fails), but a **read-only remount or a full disk would likely still let `SELECT 1` succeed**, since those only fail on *writes*. So the factories comment ("full or read-only disk surfaces as db_up=0") is a bit optimistic — `db_up` really means "the pool can serve a trivial query," not "the store is writable." If you want to catch write failures too, a tiny write probe (e.g. to a scratch table, or `PRAGMA user_version = user_version`) would be needed. Fine to leave as-is for a liveness signal; just worth aligning the comment with what `SELECT 1` actually proves.

Everything from my prior review still holds (RWO+WAL pairing correct, migrate-before-serve loud abort, optional store keeps tests clean). CI still isn't reporting on the head commit — confirm `fmt`/`clippy`/`test` + `helm lint`/`template` + the image build (bundled `libsqlite3-sys`) are green before merge. Re-approving.

Issue #183, opened 2026-04-30
Profile P2P message round-trip latency end-to-end


bagelface:
## Parent Epic
gas-killer/roadmap#9

## Context
The epic calls out P2P round-trip latency as an unknown. We currently have no measurements of how long a message takes to travel from router → node → back to router under real conditions. Without this, we cannot tell whether the gas-analyzer latency or the P2P transport is the dominant factor in overall request latency.

Key unknowns in the current setup (`commonware-p2p v0.0.63`):
- What is the actual message serialization/deserialization cost?
- What is the network transit time between router and node in the deployed topology?
- Does the 1 msg/sec rate limit (`Quota::per_second(NZU32!(1))` in `node/src/main.rs:347` and `router/src/main.rs:272`) cause queuing delay?
- Is the 256-message backlog ever saturated in practice?

## Investigation Goals
1. **Add timing instrumentation** to the P2P send/receive path:
   - Timestamp when the router sends a task message
   - Timestamp when the node receives and begins processing
   - Timestamp when the node's signature response is received by the router
   - Record these as `tracing` span fields (or log lines) so they appear in production logs

2. **Run a load scenario** using `scripts/run_scenario.rs` and capture the timing data

3. **Identify the bottleneck**: is it transport, serialization, gas-analyzer execution time, or rate limiting?

## Deliverable
Comment in this issue with:
- P99 round-trip latency under normal load (1 task at a time)
- P99 round-trip latency under burst load (N tasks queued)
- Whether the rate quota or backlog is ever the bottleneck
- Recommended follow-up tuning tasks with concrete numbers

## Files Likely Touched (instrumentation only)
- `node/src/main.rs` — add span/log around message receive and sign
- `router/src/main.rs` — add span/log around message send and response receipt
- `common/src/validator.rs` — add span around gas-analyzer call to isolate its contribution

## Unknowns / Risks
- Meaningful measurement requires a realistic network environment (not just localhost) — may need to run against the staging deployment
- `commonware-p2p` internals are not in this repo; profiling transport may require reading the library source

bagelface:
This is in the wrong repo https://github.com/gas-killer/service/issues/203

PR #295, opened 2026-06-30
refactor: migrate to commonware 2026.5.0

rubydusa:
Depends on / pairs with the restaking migration: BreadchainCoop/commonware-restaking#183 (https://github.com/BreadchainCoop/commonware-restaking/pull/183)

## Why

The service depends on the `commonware-restaking` AVS crates **and** on `commonware` directly. Once restaking moved to commonware `2026.5.0` (PR #183 above), the service's graph contained **two** versions of every commonware crate (`0.0.63` + `2026.5.0`). Because restaking's public API is expressed in commonware's types, the version mismatch broke compilation — even though restaking's own API *shape* never changed. (`commonware_codec::Read@0.0.63` ≠ `…@2026.5.0`, etc.)

The fix is to bring the service onto the same commonware version, then resolve the `0.0.63 → 2026.5.0` API breakages in the service's own code.

## What

- **deps:** `commonware-*` `0.0.63` → `2026.5.0`; `governor` `0.6` → `0.10`; dropped the direct `prometheus-client` dependency (now reached via commonware's `telemetry::metrics` re-export, so the version follows commonware automatically)
- **p2p:** `lookup::Config::local` (4 args), `attempt_unregistered_handshakes` → `bypass_ip_check` (same K8s semantics, kept `true`), `Address`-typed recipients, synchronous `oracle.track`, `with_label` → `child`
- **runtime:** `tokio::Context` is no longer `Clone` — derive a child context per `spawn`; `HealthState` holds `Arc<tokio::Context>` so the axum metrics/health handler state stays cloneable (`encode()` resolves via `Deref`; the metrics registry is shared across contexts)
- **metrics:** route prometheus types (`Counter`/`Gauge`/`Histogram`/`Registry`/`encode`) through `commonware_runtime::telemetry::metrics`

## Verification

- `cargo check --workspace --all-targets` — clean, no warnings
- `cargo test --workspace --no-run` — all test targets compile

Tests were not executed (they require live RPC / env such as `HTTP_RPC`).

> Merge after the restaking PR lands so the git dependency can be repointed off the branch.


rubydusa:
Need only to replace the branch in the cargo imports once https://github.com/BreadchainCoop/commonware-restaking/pull/183 is merged

node/src/main.rs
```diff
@@ -35,7 +35,9 @@ const APPLICATION_NAMESPACE: &[u8] = b"_COMMONWARE_AGGREGATION_";
 #[derive(Clone)]
 struct HealthState {
     ready: Arc<AtomicBool>,
-    context: tokio::Context,
+    // `tokio::Context` is no longer `Clone`; `Arc` makes the axum handler state cloneable.
```
bagelface: I don't think we should include comments like "X is no longer Y" or similar.

node/src/main.rs
```diff
@@ -315,21 +318,25 @@ fn main() {
         // where external operators are behind NAT. IP-based pre-filtering cannot work in either
         // case; authentication relies entirely on the cryptographic handshake (peer public keys
         // checked against the registered operator set), which is secure for both topologies.
-        p2p_cfg.attempt_unregistered_handshakes = true;
+        // (Renamed from `attempt_unregistered_handshakes` in commonware 2026.5.0; same
+        // semantics: skip the source-IP match so known peers can connect from unexpected IPs.)
+        p2p_cfg.bypass_ip_check = true;
```
bagelface, line 319: Similarly, this comment has too much context to this specific update.

node/src/main.rs
```diff
@@ -315,21 +318,25 @@ fn main() {
...
         // recommended() throttles peer discovery for large open gossip networks where aggressive
         // dialing is abusive. gas-killer instead runs a small, static, allowlisted operator set in a
         // full mesh: every participant dials every other, so both ends frequently dial at once and
         // one connection loses the reservation race. The loser must re-dial quickly, and an operator
         // that restarts must rejoin the signing quorum in seconds rather than ~a minute. Restore fast
-        // (re)discovery while keeping recommended's abuse-resistance (concurrent-handshake cap, subnet
-        // rate limit, ping cadence).
+        // (re)discovery (these match Config::local's values) while keeping recommended's
+        // abuse-resistance (concurrent-handshake cap, subnet rate limit, ping cadence).
+        // Note (2026.5.0): the old `query_frequency` knob was removed, and the old
```
bagelface: Similarly, this comment has too much context to this specific update.

node/src/main.rs
```diff
@@ -340,10 +347,9 @@ fn main() {
             tracing::info!(key = ?key, addr = ?addr, "oracle recipient");
         }
 
-        // Register authorized peers with the oracle
-        oracle
-            .update(0, OrderedAssociated::from_iter(recipients.clone()))
-            .await;
+        // Register authorized peers with the oracle. `track` is synchronous now and
```
bagelface: "now" is contextual to this change. Update to be more acontextual

router/src/main.rs
```diff
@@ -38,7 +38,10 @@ const APPLICATION_NAMESPACE: &[u8] = b"_COMMONWARE_AGGREGATION_";
 #[derive(Clone)]
 struct HealthState {
     ready: Arc<AtomicBool>,
-    context: tokio::Context,
+    // `tokio::Context` is no longer `Clone`, but axum requires the handler state to
```
bagelface: update comment to be more acontextual

router/src/main.rs
```diff
@@ -184,24 +185,28 @@ fn main() {
     // where external operators are behind NAT. IP-based pre-filtering cannot work in either
     // case; authentication relies entirely on the cryptographic handshake (peer public keys
     // checked against the registered operator set), which is secure for both topologies.
-    p2p_cfg.attempt_unregistered_handshakes = true;
+    // (Renamed from `attempt_unregistered_handshakes` in commonware 2026.5.0; same
+    // semantics: skip the source-IP match so known peers can connect from unexpected IPs.)
```
bagelface: update comment to be more acontextual

router/src/main.rs
```diff
@@ -184,24 +185,28 @@ fn main() {
...
     // recommended() throttles peer discovery for large open gossip networks where aggressive
     // dialing is abusive. gas-killer instead runs a small, static, allowlisted operator set in a
     // full mesh: every participant dials every other, so both ends frequently dial at once and one
     // connection loses the reservation race. The loser must re-dial quickly, and an operator that
     // restarts must rejoin the signing quorum in seconds rather than ~a minute. Restore fast
-    // (re)discovery while keeping recommended's abuse-resistance (concurrent-handshake cap, subnet
-    // rate limit, ping cadence).
+    // (re)discovery (these match Config::local's values) while keeping recommended's
+    // abuse-resistance (concurrent-handshake cap, subnet rate limit, ping cadence).
+    // Note (2026.5.0): the old `query_frequency` knob was removed, and the old
```
bagelface: update comment to be more acontextual

router/src/main.rs
```diff
@@ -261,17 +266,18 @@ fn main() {
...
             .with_max_level(tracing::Level::DEBUG)
             .with_writer(std::io::stdout)
             .finish();
         _ = tracing::subscriber::set_default(subscriber);
 
-        // Provide authorized peers
-        let authorized = OrderedAssociated::from_iter(recipients.clone());
-        oracle.update(0, authorized).await;
+        // Provide authorized peers. `track` registers a peer set (id 0) and is no
+        // longer async; `recipients` already holds `Address` values.
```
bagelface: update comment to be more acontextual

router/src/main.rs
```diff
@@ -299,8 +305,19 @@ fn main() {
         // Custom Prometheus metrics — shared with executor, creator, and ingress via builder
         let metrics = Arc::new(MetricsCollector::new());
 
+        // `tokio::Context` is no longer `Clone`, and the orchestrator builder consumes
```
bagelface: update comment to be more acontextual

bagelface (approved):
lgtm. My only issue is with the comments having too much context to this specific change. "X no longer does this", "X is now Y", "commonware update to do this". We don't want to include context like that in comments because they lose relevancy the moment this is merged in. Comments should be matter of fact and acontextual.

BreadrichEngels (approved):
This is the service half of the commonware 2026.5.0 migration paired with BreadchainCoop/commonware-restaking#183, and it's a clean, faithful adaptation of the same API changes. The code is correct; the merge is gated on two coordination items. Approving on the code with those called out.

**What's right:**
- The p2p/runtime adaptations are consistent across `node` and `router`: `Address`-wrapped recipients, synchronous `oracle.track` with `Map::from_iter_dedup` (no more clone), `with_label` → `child`, `bypass_ip_check` (kept `true`, K8s semantics preserved), `peer_connection_cooldown = 1s` as the inverse of the old `1/s` per-peer rate, `MAX_MESSAGE_SIZE` as `u32`. These line up with #183's counter-example changes.
- **The `tokio::Context`-no-longer-`Clone` handling is done correctly.** In `router/src/main.rs` you pre-derive every child context (`executor_ctx`, `speculative_ctx`, `orchestrator_task_ctx`, `healthz_ctx`, `health_ctx`) *before* moving the root `context` into `OrchestratorBuilder::new(context, …)`, which consumes it — that's the right ordering to avoid a use-after-move, and wrapping `HealthState.context` in `Arc` keeps the axum state `Clone`.
- **Observability is preserved:** child contexts share the one runtime metrics registry, so `/metrics` encoded from `health_ctx = context.child("metrics")` still reports the orchestrator/network families registered under sibling scopes. That's the property that would silently break here, and the derive-before-move keeps it intact — worth confirming on the soak that the full metric set still shows up.

**🚫 Merge gate 1 — the git dependency is branch-pinned.** `Cargo.toml` points the `commonware-avs-*` crates at `branch = "9-research-spike-commonware-library-update"` (the #183 branch). Before this merges, #183 must land and this must be repointed to `main` with `Cargo.lock` bumped to the #183 merge commit — otherwise the service pins to a feature branch that can move or disappear. You and rubydusa both note this; flagging it as the hard gate. Order: merge #183 → repoint `Cargo.toml`/`Cargo.lock` here → merge #295.

**🚫 Merge gate 2 — coordinated full-fleet deploy.** As on #183: the commonware-p2p protocol change means 0.0.63 and 2026.5.0 peers can't transport with each other, so the router and **all nodes** must roll to this together — not a mixed-version rollout. This is a deploy-time constraint, but it needs to be in the plan since a partial rollout would stall round formation entirely.

**Notes:**
- Tests weren't run (need live RPC), only `cargo check` + `test --no-run`. Given this touches p2p, runtime, and metrics, a **testnet soak is the real gate**: verify intra-cluster connectivity (the `bypass_ip_check` rename and `peer_connection_cooldown` behaving as the old rate), round formation end-to-end, and that `/metrics` still emits every family.
- Same comment-context nit bagelface raised on both PRs — the "(Renamed from … in commonware 2026.5.0)" / "Note (2026.5.0): …" inline comments will age poorly on `main`. #183 already carries the durable history in `docs/commonware-migration-2026.5.0.md`; trim these to acontextual statements (or point at that doc).
- CI isn't reporting on the head commit, and this is a sizable dep bump (governor 0.10, dropped direct `prometheus-client`) — confirm the workspace + image build are green before merge.

Approving the code; just land #183 first, repoint the dep, and deploy the fleet together.

Cargo.toml
```diff
@@ -24,17 +24,17 @@ async-trait = "0.1"
...
-commonware-cryptography = "0.0.63"
-commonware-macros = "0.0.63"
-commonware-p2p = "0.0.63"
-commonware-runtime = "0.0.63"
-commonware-utils = "0.0.63"
+commonware-avs-bindings = { git = "https://github.com/BreadchainCoop/commonware-restaking", branch = "9-research-spike-commonware-library-update" }
+commonware-avs-core = { git = "https://github.com/BreadchainCoop/commonware-restaking", branch = "9-research-spike-commonware-library-update" }
+commonware-avs-eigenlayer = { git = "https://github.com/BreadchainCoop/commonware-restaking", branch = "9-research-spike-commonware-library-update" }
+commonware-avs-node = { git = "https://github.com/BreadchainCoop/commonware-restaking", branch = "9-research-spike-commonware-library-update"}
+commonware-avs-router = { git = "https://github.com/BreadchainCoop/commonware-restaking", branch = "9-research-spike-commonware-library-update"}
```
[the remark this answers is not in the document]
rubydusa: @bagelface wdyt? I think it's a good idea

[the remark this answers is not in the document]
bagelface: We've just been pinning to default branch. The Cargo.lock file pins it to a rev. I guess this is more explicit, just ugly. Idk i could take it or leave it.

bagelface:
Please scrub Claude as an author from the commits. We don't have any Claude commits in the main branch and I think it looks better that way.

Cargo.toml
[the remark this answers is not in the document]
rubydusa: I'll be removing the branch on this once the other PR will get merged

PR #296, opened 2026-07-01
feat: token-based API key authentication with revocation

### LINKED ISSUE #188: Token-based API key authentication with revocation

bagelface:
## Summary

Implements token-based API key authentication with revocation for the router ingress. Replaces the single shared `INGRESS_PASSWORD` with per-client, individually revocable API keys, managed through an admin API and persisted in the SQLite store.

Addresses #188. Part of the v1-beta epic gas-killer/roadmap#15.

> **Stacked on #294** (`bagelface/scaffold-sqlite-store`, the SQLite scaffold for #293). Until #294 merges, this PR's diff includes those commits; it self-cleans to just the API-key commit once #294 lands. Review the last commit (`feat: add token-based API key authentication with revocation`) for this change in isolation.

## What's included

**Persistence** (`router/src/store/api_keys.rs`, `migrations/0001_api_keys.sql`)
- `api_keys` table storing only a **keccak-256 hash** of each key. The raw `gk_<32-byte hex>` value is returned once at creation and never persisted, so a database leak cannot recover a live key. Revocation is a soft delete (`revoked_at`), preserving an audit trail.
- `create_api_key` / `list_api_keys` / `revoke_api_key` / `verify_api_key`. Verification is a single `UPDATE … WHERE key_hash = ? AND revoked_at IS NULL RETURNING id`, authenticating and stamping `last_used` atomically.

**HTTP API** (`router/src/ingress.rs`)
- `POST /admin/keys` → `201 {id, key, label, created_at}` — key shown exactly once.
- `GET /admin/keys` → active key metadata (`id`, `label`, `created_at`, `last_used`); never the key value or hash.
- `DELETE /admin/keys/:id` → `204`, or `404` if no active key with that id.
- Admin endpoints guarded by a separate `ADMIN_KEY` secret (constant-time compare); `503 NOT_CONFIGURED` when unset so a misconfigured operator gets a clear signal instead of an open door.
- `/trigger` now authorizes via a valid, unrevoked API key.

**Config / deploy** — `ADMIN_KEY` wired through `example.env`, Helm `values.yaml`, `secret.yaml`, `router-deployment.yaml` (injected only when set), and documented in the chart README.

## Acceptance criteria

- [x] `POST /admin/keys` creates a key and returns `{id, key, label, created_at}`; key shown once
- [x] `DELETE /admin/keys/{id}` revokes immediately; subsequent requests using it return 401
- [x] `GET /admin/keys` lists active key metadata, never the key value
- [x] Task requests require a valid, unrevoked key via `Authorization: Bearer <key>`
- [x] Admin endpoints protected by a separate `ADMIN_KEY` credential
- [x] Key store survives router restarts (backed by the #293 SQLite store)
- [x] Constant-time comparison preserved for shared secrets

## Design notes

- **keccak-256 hash-lookup, not a slow password hash.** Keys carry 256 bits of entropy, so preimage resistance — not argon2-style work — is what defeats reversal and timing attacks; the raw secret is never compared byte-wise. Constant-time comparison is preserved for the shared secrets (`ADMIN_KEY`, legacy password), which are the values compared directly.
- **Endpoint is `/trigger`, not `/tasks`** (the issue's name). The requirement is applied to the real task-submission endpoint rather than renaming it, which would break clients, e2e, and docs.
## Testing

- `cargo fmt --check`, `clippy --all-targets --all-features -D warnings`, and full `cargo test` all clean. Router tests 84 → 102 (+18: 8 store, 10 ingress/admin covering admin auth, create/list/revoke, and key acceptance/revocation on `/trigger`).
- Helm renders and lints; `ADMIN_KEY` appears in the secret and router env only when configured.
- `helm-integration-tests.yml` gains a self-contained admin-flow step (auth-required → mint → list-without-leak → key passes the gate → revoke → revoked-key 401) against the real deployed router and persisted store, without disturbing the on-chain trigger flow.


BreadrichEngels (changes requested):
The auth design is genuinely strong — this is careful work. Storing only the keccak-256 hash (raw `gk_<64 hex>` shown once, 256-bit entropy, so hash-lookup rather than argon2 is the right call), the unified `invalid_at` for expiry+revocation with a single `IS NULL OR > now` check, the atomic verify-and-stamp `UPDATE … RETURNING id`, random non-sequential ids, parameterized queries throughout, the admin API failing **closed** (503 `NOT_CONFIGURED` when `ADMIN_KEY` is unset, constant-time compare when set), and listings that structurally can't leak the key or hash — all correct, and the 18 tests cover the paths that matter (revoked/unknown/expired keys rejected on `/trigger`). Nice.

But there's one blocking issue: **the PR description and the code disagree about `INGRESS_PASSWORD`, in a security- and rollout-critical way.**

**🚫 `INGRESS_PASSWORD` is not "retained as a legacy fallback" — it's removed entirely.** The summary says three times that it's kept as a constant-time fallback "so existing single-token deployments and the e2e harness keep working during the beta," and lists "constant-time comparison preserved for shared secrets (…legacy password)" as met. But `authorize_task_request` only calls `store.verify_api_key(token)` — there is no password check — and the diff deletes the `password` field from `IngressState`, the `env::var("INGRESS_PASSWORD")` read in `factories`, the old `/trigger` password branch, the Helm `INGRESS_PASSWORD` secret/deployment env, and the CI `Authorization: Bearer $CI_INGRESS_PASSWORD` trigger call. So the fallback described in the PR does not exist.

Why this matters beyond a stale description: because the store is *always* present in a real deployment (#294's `factories` opens it and aborts startup on failure), the moment this ships **every `/trigger` request requires a valid minted API key** — there is no gradual beta migration, it's a hard cutover. Any client, the e2e/scenario harness, or internal caller still sending `INGRESS_PASSWORD` (or no token) gets a 401 across the board. Please either:
1. **Correct the description and plan the cutover explicitly** — state that `INGRESS_PASSWORD` is *removed* (breaking), document the migration (set `ADMIN_KEY`, mint a key per client), and confirm every deployed trigger path — the on-chain e2e flow, the scenario scripts, and any live testnet clients — is re-keyed *before* this deploys. (The CI admin-flow migration suggests this was the intent; it just needs to be said, and the trigger paths verified.) Or
2. **Actually implement the fallback** if that was the goal — add a constant-time `INGRESS_PASSWORD` check in `authorize_task_request` — though the wholesale deletion of the plumbing suggests (1) is what you meant.

Right now a reader (or an operator planning the upgrade) is actively misled about whether their existing token keeps working. It won't.

**Also fix — doc-comment got split:** `fn unix_now()` was inserted *inside* `CreateApiKeyRequest`'s doc comment, so `/// Request body for POST /admin/keys … an empty body creates an` now attaches to `unix_now`, and `/// unlabeled key. invalid_at …` attaches to the struct. Both rendered docs are garbled — move `unix_now` (and its own doc) out of the middle of the struct's doc block.

**Minor (non-blocking, defense-in-depth):**
- **Fail-open when `store` is `None`.** `authorize_task_request` returns `Ok(())` with no store, i.e. `/trigger` is fully open. Prod always has a store so it's safe today, and it matches the old "open when unconfigured" posture — but it's a latent footgun: any future code path that builds `IngressState` without `with_store` silently disables auth. Consider fail-closed with an explicit dev-mode opt-in (env flag) rather than implicit-open. Contrast with the admin API, which you correctly made fail-closed — worth the same treatment here.
- **Auth runs a DB query before the capacity check**, and `verify_api_key` is an `UPDATE` (writes `last_used`) on every *valid* request. Fine at beta load with WAL + the 5s busy_timeout, but a credential-stuffing flood incurs a keccak + indexed lookup per request ahead of load-shedding, and steady valid traffic serializes on the single SQLite writer for the `last_used` stamp. Worth keeping in mind if ingress QPS grows (e.g. throttle `last_used` updates).

Stacked on #294, so as you note, review/merge the API-key commit after the scaffold lands. CI isn't reporting on the head commit yet — get `fmt`/`clippy`/the +18 tests and the helm admin-flow step green before merge.

Happy to re-approve once the `INGRESS_PASSWORD` story is reconciled (description + cutover confirmed, or the fallback implemented) and the doc block is fixed — the underlying auth implementation itself I'm satisfied with.

BreadrichEngels (approved):
Re-review — the blocker is resolved, and you addressed more than I flagged. Clearing the change request.

**✅ The `INGRESS_PASSWORD` story is now consistent — and, importantly, you implemented the cutover, not just re-worded it.** The description's "retained as a fallback / beta migration" claims are gone (removal is intentional), and the deployed trigger paths are actually re-keyed: the CI helm-integration flow mints a working key through the admin API and authenticates the on-chain `/trigger` calls with `Bearer $GAS_KILLER_API_KEY` (the old `$CI_INGRESS_PASSWORD` trigger call is deleted), and `send_request` / `run_e2e` / the scenarios now read `GAS_KILLER_API_KEY`. That's exactly the "confirm every deployed trigger path is re-keyed before this ships" I asked for — done in-band. Operators still need to set `ADMIN_KEY` and mint keys for any *external* clients before rollout, but the repo's own paths are handled and the docs now say so.

**✅ Fail-open → fail-closed.** You went past my defense-in-depth note: `/trigger` now *rejects* every request when no store is configured, with an explicit `allow_unauthenticated` opt-in reserved for store-less test harnesses (and documented "never use in production"). That removes the latent footgun where a store-less `IngressState` would silently disable auth — the task path now matches the admin API's fail-closed posture. Nice.

**✅ Doc-comment split fixed** — `CreateApiKeyRequest`'s doc is contiguous and `unix_now` is no longer wedged inside it.

The core auth implementation I was already satisfied with (hash-only storage, unified `invalid_at`, atomic verify-and-stamp, constant-time `ADMIN_KEY`, fail-closed admin API, no key/hash in listings) is unchanged. 

**Two tiny description tidy-ups (non-blocking, docs only):**
- The Design Notes constant-time bullet still says "shared secrets (`ADMIN_KEY`, legacy password)" — there's no legacy password anymore, so it's just `ADMIN_KEY`.
- The Persistence bullet still describes revocation as a soft delete on `revoked_at` (`WHERE key_hash = ? AND revoked_at IS NULL`), but the schema uses the unified `invalid_at`. Worth aligning so the description matches the migration.

`mergeable_state` is `blocked` (my prior change-request; this approval clears it) and CI still isn't reporting — get `fmt`/`clippy`/the 100+ tests and the helm admin-flow step green before merge, and land after #294 as noted. Approving.

BreadrichEngels (approved):
Follow-up on the two new commits (`e5488c7`, `14043e3`) — both are good hardening, and they complement each other well.

**`e5488c7` — ingress path allowlist. Strong defense-in-depth.** Switching the K8s Ingress from an implicit catch-all to an explicit allowlist (`publicPaths`, default `[/trigger, /avs-metadata, /healthz]`) means `/admin/*` is no longer internet-facing — the admin key-management API is now reachable only in-cluster, on top of its `ADMIN_KEY` guard. That's exactly the right posture: minting/revoking keys shouldn't be exposed to the public even behind a shared secret. Details check out:
- Correctly excludes `/admin`, `/metrics`, and `/readyz` from public exposure (metrics especially shouldn't be public; readiness is kubelet-probed in-cluster).
- Falls back to the safe list "never to a catch-all `/`", so a misconfigured `publicPaths` can't accidentally re-expose everything.
- The in-cluster Service still exposes all ports, so the helm-integration admin-flow test (which hits the router via port-forward/Service, not the Ingress) keeps working — the allowlist only constrains the public edge.

**`14043e3` — ship `create_api_key` in the router image. Correct complement.** With `/admin` no longer public, operators mint keys from inside the cluster, so bundling the CLI in the image is the right enabler (`kubectl exec <router-pod> -- create_api_key --url http://localhost:8080 …`). `cargo build -p gas-killer-router` builds the `src/bin/*` targets, so the `cp`/`COPY` of `create_api_key` (and `generate_key`) resolve, and shipping it doesn't widen the attack surface — anyone with pod-exec already has `ADMIN_KEY` from the env and `curl`.

**One thing to confirm in the docs (minor):** `create_api_key`'s `--env prod`/`--env testnet` modes target the *public* URLs (`https://api.gaskiller.xyz`, etc.), but `/admin/keys` is no longer routed there — so those modes will now 404 against the public edge, and admin ops must go through `--url http://localhost:8080` in-cluster (or a port-forward). Since `e5488c7` already updated `README.md`/the chart README, just make sure they steer operators to the in-cluster `--url` path and don't leave `--env prod` looking like the way to mint against production, or it's a confusing dead end.

No code-behavior change to the auth path itself here, and everything from my prior approval still stands. `mergeable_state` is `unstable` / CI not reporting — get the build (now producing the extra bins) + tests + helm steps green, and land after #294. Approving.

BreadrichEngels (approved):
Follow-up on the two new commits — both good, and together they close out the API-key rollout coherently.

**`bc593649` — `create_api_key` defaults to in-cluster.** This fully resolves the `--env`/edge-404 note from my last review. `resolve_base` now returns `http://localhost:8080` when no flags are given, `--env` lost its `testnet` default and is opt-in with help that spells out "admin is private by default, prefer `--url`," and the module docs show the `kubectl exec` / port-forward workflow and explicitly warn that `--env` only works if `/admin` is added to `ingress.publicPaths`. The new `base_resolution_defaults_to_in_cluster` test pins the precedence (url > env > in-cluster default). Exactly right — the tool now points operators at the reachable, private endpoint instead of a public URL that won't route `/admin`.

**`79ce4ab5` — scenario configs carry an API key.** The `Config.api_key` field with a config-first, `GAS_KILLER_API_KEY`-env-fallback resolution is clean, threaded correctly through both serial and parallel paths (`Arc<Option<String>>` → `as_deref()`), and `send_request` only attaches `Authorization: Bearer` when the key is non-empty, so the no-auth/dev path still works. Good hygiene: the startup line logs `API key: configured|none` rather than the value, so the secret never lands in scenario output/CI logs. This completes the cutover — the load/smoke scenarios that hit `/trigger` will now authenticate against an auth-required router instead of 401ing.

**One trivial check:** the per-scenario `.toml` additions should reference the env (e.g. `api_key = "$GAS_KILLER_API_KEY"`, resolved by your `interpolate_env`) rather than a literal key — just confirm no real token got committed into a checked-in scenario file. Given the env-fallback design it's almost certainly a `$VAR` reference; flagging only because these files are in the repo.

(Also nice to see `parse_call_data` now reports "malformed byte array" on a missing `]` instead of falling through to a hex error — that was the exact nit from #289.)

No behavioral concerns. CI still isn't reporting on the head commit — get the build/tests/helm steps green, and land after #294. Approving.

PR #297, opened 2026-07-01
fix: clean up spurious CI diagnostics in e2e and helm tests

bagelface:
Two CI test workflows produced misleading output that didn't reflect any real problem with the service. Both fixes are CI-only — no changes to router/executor behavior.

## e2e-test: spurious `cast run` OutOfGas
The post-success diagnostic ran `cast run <tx>` to print a trace. `cast run` re-simulates the transaction using its recorded gas limit, and the router submits `verifyAndUpdate` with alloy's bare `eth_estimateGas` result — which on the anvil fork equals the exact gas consumed (`gas_used == gas_limit`, zero headroom). The re-simulation needs a hair more than the live execution, so it always reported `OutOfGas` / `Error: Transaction failed` for a transaction that actually succeeded (`status=true`).

- **On success**, the trace now comes from `cast rpc debug_traceTransaction <tx> '{"tracer":"callTracer"}'`, which reflects the real mined execution instead of re-simulating it — so it shows the true (successful) call tree with no spurious OutOfGas.
- **On failure**, `cast run` runs in the failure branch, where a genuinely reverted transaction traces faithfully and surfaces the real revert reason.

## helm-integration-tests: SumCalculated events always counted as 0
The event check reported `Found 0` and threw `integer expression expected`, from three compounding bugs (confirmed against a Sepolia-fork anvil):

- `--topic` is not a valid `cast logs` flag in the current foundry (topic is positional), so the command errored out entirely and was swallowed by `2>/dev/null`.
- `--from-block 0` on a fork is forwarded to the upstream RPC, which rejects `getLogs` over more than 50000 blocks. Now bounded to the block captured just before the triggers.
- `grep -c … || echo "0"` produced `"0\n0"` on zero matches. Counting now uses `cast logs --json | jq 'length'` with an empty-array guard.

Verified end to end: 3 emitted events → count 3; zero matches → 0; RPC error → 0.

### DIFF
--- a/.github/workflows/helm-integration-tests.yml
+++ b/.github/workflows/helm-integration-tests.yml
@@ -412,6 +412,9 @@ jobs:
           INITIAL_COUNT=$(cast call "$ARRAY_SUMMATION_ADDRESS" "stateTransitionCount()(uint256)" --rpc-url http://localhost:8545 | tr -d '[:space:]')
           echo "Initial stateTransitionCount: $INITIAL_COUNT"
 
+          # Block height before triggering, used to bound the SumCalculated event query below.
+          START_BLOCK=$(cast block-number --rpc-url http://localhost:8545)
+
           NUM_TRIGGERS=3
           SUCCESSFUL=0
 
@@ -447,9 +450,10 @@ jobs:
             exit 1
           fi
 
-          # Check for SumCalculated events
+          # Count SumCalculated events emitted since the triggers began.
           EVENT_TOPIC=$(cast keccak 'SumCalculated(uint256,uint256)')
-          EVENT_COUNT=$(cast logs --from-block 0 --address "$ARRAY_SUMMATION_ADDRESS" --topic $EVENT_TOPIC --rpc-url http://localhost:8545 2>/dev/null | grep -c "blockNumber" || echo "0")
+          EVENT_JSON=$(cast logs --from-block "$START_BLOCK" --address "$ARRAY_SUMMATION_ADDRESS" "$EVENT_TOPIC" --rpc-url http://localhost:8545 --json 2>/dev/null || echo '[]')
+          EVENT_COUNT=$(echo "$EVENT_JSON" | jq 'length')
           echo "Found $EVENT_COUNT SumCalculated events"
 
           if [ "$EVENT_COUNT" -ge "$NUM_TRIGGERS" ]; then
--- a/scripts/run_e2e_test.sh
+++ b/scripts/run_e2e_test.sh
@@ -233,6 +233,13 @@ else
     docker compose logs --tail=100 router || true
     echo -e "${YELLOW}Recent node logs:${NC}"
     docker compose logs --tail=50 node-1 node-2 node-3 || true
+    # Trace the verifyAndUpdate transaction, if one was submitted, to surface a revert reason.
+    # cast run re-simulates the transaction, so this is best-effort diagnostic output.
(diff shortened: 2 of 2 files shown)


BreadrichEngels (approved):
Clean, correctly-diagnosed CI fix — test tooling only, no production impact. Each of the three helm-test bugs and the e2e change check out:

- **`--topic` → positional topic.** Right — `cast logs` takes the event sig/topic positionally, so `--topic` errored and was swallowed by `2>/dev/null` (hence the silent `Found 0`). Passing the `cast keccak` topic0 hash positionally is valid.
- **`--from-block 0` → `$START_BLOCK`.** Correct fix for the fork's >50k-block `getLogs` rejection, and capturing `START_BLOCK` *before* the triggers still covers the events emitted during them.
- **`grep -c … || echo "0"` → `--json | jq 'length'`.** This was the actual "integer expression expected" culprit (`grep -c` prints `0` *and* exits non-zero on no match, so `|| echo "0"` appended a second line → `"0\n0"`). The `cast … --json 2>/dev/null || echo '[]'` + `jq 'length'` gives a clean integer on all three paths (3 events → 3, none → 0, RPC error → `[]` → 0). Matches your end-to-end verification.
- **`cast run --quick`** replays only the target tx against the parent state, which is the correct way to kill the spurious `OutOfGas` on an already-successful tx; it's a diagnostic trace behind `|| echo`, so zero risk.

**One tiny defensive note (non-blocking):** the `|| echo '[]'` guards a cast *failure* (non-zero exit), but if `cast logs --json` ever exits 0 with empty stdout instead of `[]`, `jq 'length'` would error on empty input. Your verification shows the installed foundry emits `[]` for zero matches, so this is fine as-is — just something a `${EVENT_JSON:-[]}` or `jq 'length' 2>/dev/null || echo 0` would harden if foundry's output ever changes.

Since this is itself a CI change, the real confirmation is the e2e + helm-integration workflows going green on this branch — worth a run before merge (status isn't reporting yet). Approving.

BreadrichEngels (approved):
Re-review — and the e2e rework is actually a better fix than the `--quick` I'd suggested, because it addresses the real root cause. Dropping the `|| echo '[]'` nit as you asked.

You correctly diagnosed *why* `cast run` reported OutOfGas: the router submits `verifyAndUpdate` with a bare `eth_estimateGas` result, so on the fork `gas_limit == gas_used` (zero headroom), and *any* re-simulation — including `cast run --quick` — needs a hair more than the live run and trips OutOfGas. So `--quick` wouldn't have reliably fixed it; switching the success path to `cast rpc debug_traceTransaction … callTracer` is the right move — it traces the *already-mined* execution instead of re-simulating, so it can't spuriously revert. Keeping `cast run` on the failure branch is also correct: a genuinely reverted tx re-simulates faithfully and surfaces the revert reason, which is exactly what you want when debugging a failure. Both are behind `|| true`, so purely diagnostic.

Mechanics check out: `cast rpc debug_traceTransaction <hash> '{"tracer":"callTracer"}'` has the right positional param arity, anvil supports the callTracer, and `| jq '.' || true` stays non-fatal even under pipefail. The `[ -n "$TX_HASH" ] && command -v cast` guards handle the missing-hash / no-cast cases cleanly.

The helm-integration change is unchanged from my last pass and still correct (positional topic, `$START_BLOCK` bound, `--json | jq 'length'`).

As before, the real confirmation is the e2e + helm-integration workflows going green on this branch — status still isn't reporting, so run them before merge. Re-approving.

Issue #194, opened 2026-05-05
Task status polling endpoint


bagelface:
## Parent Epic
gas-killer/roadmap#15

## Background

Once tasks have IDs (https://github.com/gas-killer/service/issues/193), users need a way to check on them. This is the primary interface between users and the async aggregation process.

## Acceptance Criteria

- [ ] `GET /tasks/{task_id}` returns the full task state
- [ ] When status is `failed` or `expired`, the response includes a non-empty `error` string
- [ ] `GET /tasks` returns a paginated list scoped to the caller's API key, ordered by `created_at` descending; supports `?status=` filter
- [ ] Returns `404` for unknown task IDs; `403` if the requesting key does not own the task

**Payload note:** the response shape below always includes a `payload` field, but *populating* it when `status == "ready"` is tracked entirely in gas-killer/service#195 (TASK-3) — this issue is not blocked by it. `payload` stays `null` until #195 lands; #195's own acceptance criteria cover updating this endpoint's response once the payload exists.

## Response Shape

```json
{
  "task_id": "550e8400-e29b-41d4-a716-446655440000",
  "status": "queued | processing | ready | failed | expired",
  "created_at": 1777291200,
  "updated_at": 1777291245,
  "error": null,
  "payload": null
}
```

`payload` is populated when `status == "ready"` (see #195 / TASK-3).

`created_at` and `updated_at` are unix timestamps in seconds, consistent with the rest of the API (e.g. the admin key listing). The ISO-8601 example previously shown here was illustrative only.

## Metadata

| Field | Value |
|-------|-------|
| Epic | Task Lifecycle Redesign |
| Story ID | TASK-2 |
| Complexity | Medium |
| Priority | P0 — must-ship |

bagelface:
Already fully implemented on `task-lifecycle-refactor` via PR #307 (merged 2026-07-03), and it's survived every merge since (the `main` reconciliation, the sequencer migration, PR #324). Re-ran the full `ingress::` test suite fresh — all 88 tests pass, including the #194-specific cases:

- `GET /tasks/{task_id}` returns full state (`get_task_returns_owner_task`)
- Non-null `error` on `failed` (`get_task_failed_includes_error`)
- `404` unknown / `403` non-owner (`get_unknown_task_returns_404`, `get_task_owned_by_other_key_returns_403`)
[...]

PR #300, opened 2026-07-03
feat: add persistent task model and store

### LINKED ISSUE #192: Persistent task store (survive restarts)

bagelface:
## Summary

Adds the persistent task model — the SQLite schema, the `Task` record, and the store CRUD — as the foundation of the async task lifecycle. This is the data-layer slice of the persistent task store (gas-killer/service#192) and the precursor to the endpoint work: `POST /tasks` (gas-killer/service#193) and the status/polling endpoints (gas-killer/service#194), which will consume these store methods.

Part of the Service v1 Beta epic (gas-killer/roadmap#15).

## What's included

- **Migration `0002_tasks.sql`** — a `tasks` table keyed by UUID, foreign-keyed to `api_keys(id)`, with a `status` CHECK constraint (`queued | processing | ready | failed | expired`), the full submitted request (target/from addresses, call data, transition index, value, block height), `payload`/`error` outputs, and `created_at`/`updated_at`. Indexes cover the per-key listing and the startup re-queue scan.
- **`store/tasks.rs`** — `TaskStatus` enum, `Task` record, and store methods: `create_task`, `get_task`, `list_tasks_for_key` (scoped, status-filtered, paginated), `update_task_status`, `mark_task_ready`, `mark_task_failed`, and `incomplete_tasks` (for restart re-queue). The full request is persisted so a task can be rebuilt and re-enqueued after a pod restart.

## Scope

Data layer only — the store methods are not wired to any endpoint in this PR. They are re-exported from the crate root and covered by tests so they compile clean under `-D warnings` ahead of the endpoint PRs. Durability guarantees (write-before-ack, startup re-queue wiring) and the Helm PVC mount from #192 are intentionally left to follow-on work once the endpoints exist.

## Testing

- 14 new unit tests for the task store (creation, FK enforcement, request round-trip, ownership-scoped listing, status filter, pagination, status transitions, restart re-queue query, status encoding).
- `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings` (clean), and `cargo test --all-targets --all-features` all pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4050,6 +4050,7 @@ dependencies = [
  "tracing",
  "tracing-subscriber 0.3.23",
  "url",
+ "uuid 1.22.0",
 ]
 
 [[package]]
@@ -9882,6 +9883,7 @@ version = "1.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a68d3c8f01c0cfa54a75291d83601161799e4a89a39e0929f4b0354d88757a37"
 dependencies = [
+ "getrandom 0.4.2",
  "js-sys",
  "wasm-bindgen",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,3 +64,4 @@ tower-http = { version = "0.5", features = ["cors"] }
 tracing = "0.1.41"
 tracing-subscriber = "0.3.19"
 url = { version = "2.5.4", features = ["serde"] }
+uuid = { version = "1", features = ["v4"] }
--- a/router/Cargo.toml
+++ b/router/Cargo.toml
@@ -89,6 +89,7 @@ tower-http = { workspace = true }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
 url = { workspace = true }
+uuid = { workspace = true }
 
 [dev-dependencies]
 tempfile = { workspace = true }
--- a/router/migrations/0002_tasks.sql
+++ b/router/migrations/0002_tasks.sql
@@ -1,0 +1,38 @@
+-- Task records tracking each submitted aggregation request through its lifecycle.
+--
+-- A task enters in the `queued` state when a request is accepted, advances to `processing`
+-- while the router aggregates operator signatures, and settles in a terminal state: `ready`
+-- (carrying an executable `payload`), `failed`, or `expired` (each carrying a human-readable
+-- `error`). Persisting the request parameters alongside the state lets the router rebuild and
+-- re-enqueue tasks still `queued` or `processing` after a restart, so an in-flight request is
+-- never silently lost when the pod recycles.
+--
+-- `key_id` scopes a task to the API key that created it, both for ownership checks on the
+-- status endpoints and so a key's own tasks can be listed. The listing filters and paginates
+-- by `(key_id, created_at)`, which the index below serves.
+--
+-- Ethereum values are stored as text: addresses in their `0x` hex form and `value` as a
+-- decimal string, since a 256-bit integer does not fit SQLite's signed 64-bit INTEGER.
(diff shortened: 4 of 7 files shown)


BreadrichEngels (approved):
Solid foundation for the task lifecycle — the schema and store CRUD are clean, correct, and thoroughly tested (14 tests covering FK enforcement, request round-trip, scoped/filtered/paginated listing, every transition, and the restart scan). Good scoping too: data-layer only, `pub` methods re-exported so they compile under `-D warnings` without being wired yet. LGTM.

Things I checked that hold up:
- **Ethereum types stored as text is the right call** — `value` as a decimal `U256` string (a 256-bit int can't fit SQLite's signed `INTEGER`), addresses as `0x` hex, `call_data` as `BLOB`. The `TryFrom<TaskRow>` parses each back with per-field error context, and `get_round_trips_request_fields` / `auto_transition_index_round_trips_as_none` confirm the round-trip including `NULL ↔ None` for auto index.
- **FK actually enforced.** `key_id REFERENCES api_keys(id)` only bites because #294's connect options set `PRAGMA foreign_keys=ON`; `create_rejects_unknown_key` verifies it. Migration order is fine (0001 creates `api_keys` before 0002 references it), and since keys are soft-deleted (`invalid_at`), never hard-deleted, there's no dangling-FK / cascade concern.
- **The reused numbered param is correct.** `WHERE key_id = ?1 AND (?2 IS NULL OR status = ?2)` binds `status.map(as_str)` once as `?2` and references it twice — valid for SQLite numbered params, and `list_filters_by_status` exercises both the filtered and null-filter paths. Indexes match the two access patterns (`(key_id, created_at DESC)` for listing, `(status)` for the re-queue scan), and `created_at DESC, id` gives a deterministic tiebreak.
- **UUID v4** via `uuid`'s `v4` (getrandom-backed); the test asserts `get_version_num() == 4`.

**One gap worth noting for the follow-on (non-blocking here):** `Expired` is a terminal state the schema/doc say "carries an `error`," but there's no `mark_task_expired(id, error)` analogous to `mark_task_failed` — the only way to reach it is `update_task_status(id, Expired)`, which doesn't record an error. So the "expired carries an error" contract isn't yet expressible. Fine to leave for the expiry-wiring PR, just flagging so it isn't forgotten there.

**Minor:** `incomplete_tasks` loads all `queued`/`processing` rows unbounded — fine at beta volume and appropriate for a one-shot startup scan, but if the interrupted backlog can get large, the re-queue PR may want to page it.

As the PR says, the durability guarantees (write-before-ack, startup re-queue wiring) and the PVC mount are the parts that make this *matter* operationally — this is just the store, and it's a good one. CI isn't reporting on the head commit yet; confirm `fmt`/`clippy`/the 14 tests are green before merge. Approving.

Issue #195, opened 2026-05-05
User-executable payload output (replace on-chain execution)


bagelface:
## Parent Epic
gas-killer/roadmap#15

## Background

Currently the router holds a funded ECDSA key and calls `verifyAndUpdate` on-chain directly. For v1 beta, we shift execution to the user: the service computes the BLS-aggregated result and returns a **ready-to-sign transaction request** that the user signs and sends themselves. This eliminates the router's need to custody a funded key and gives users full control over gas parameters and submission timing.

The completed round is persisted as a **structured bundle** (the `verifyAndUpdate` argument components), not as assembled calldata. The transaction request returned to the user is one *rendering* of that bundle — keeping the stored unit reusable (re-aggregation, or a future AA/bundler submission tier) and the wire shape forward-compatible.

gas-killer/service#194 already ships `GET /tasks/{id}` with a `payload` field in its response shape, always `null` for now — #194 was explicitly scoped to not block on this issue. This issue is what makes `payload` non-null: the polling endpoint's response shape does not change, only what populates that field.

## Acceptance Criteria

- [ ] When an aggregation round completes, the task transitions to `ready` and `GET /tasks/{id}` (#194) returns the populated `payload` object instead of `null`
- [ ] The completed round is persisted as a structured bundle — `(msgHash, quorumNumbers, referenceBlockNumber, storageUpdates, transitionIndex, targetFunction, nonSignerStakesAndSignature, chainId)` — keyed by task id; the returned `payload` is rendered from it
- [ ] `payload` is a full transaction request — `{ to, data, value, chain_id }` — that the user signs and sends **as-is, with no modification**. It is **not** bare `verifyAndUpdate` calldata: `to` and `value` are server-controlled so that adding an on-chain protocol fee (target `payable`+`fee`, or a `GasKillerEntrypoint`) or a future AA tier changes the server output, never the integrator's client code
- [ ] `valid_until_block` is the block reference used during aggregation plus a configurable grace buffer (env var `PAYLOAD_BLOCK_BUFFER`, default: 50 blocks)
- [ ] A stale or already-consumed bundle (transitionIndex no longer matches, or past `valid_until_block`) returns a typed "re-request" error rather than a payload that would revert on submission
- [ ] The `Executor` is disconnected from the aggregation completion handler — the bundle is written to the task store instead of being broadcast
- [ ] The `Executor` component is retained in the codebase for future use (e.g. account abstraction / ERC-4337 bundler submission), consuming the same persisted bundle
[...]

Issue #197, opened 2026-05-05
Task deduplication


bagelface:
## Parent Epic
gas-killer/roadmap#15

## Background

A client that retries after a timeout may submit the same logical request twice. Both tasks would pass validation, but only the first to reach the chain succeeds — the second fails because `transition_index` is already consumed. Deduplication makes retries safe and idempotent.

## Acceptance Criteria

- [ ] On `POST /tasks`, if an in-flight (`queued` or `processing`) or `ready` task exists for the same `(key_id, target_address, transition_index)`, return the existing task ID with `200 OK` instead of creating a new task
- [ ] If the existing task is `failed` or `expired`, allow re-submission (creates a new task with a fresh ID)
- [ ] The deduplication check and task creation are atomic
- [ ] The response body for a deduplicated request is identical in shape to a fresh submission, with an additional `"deduplicated": true` field

## Metadata

| Field | Value |
|-------|-------|
| Epic | Reliability & Persistence |
| Story ID | RELY-2 |
| Complexity | Low |
| Priority | P1 — strong recommendation |

bagelface:
Follow-on note from the async submission PR (#302): the submit path persists the task (`create_task`) *then* enqueues it. If `sender.send` fails after the row is written (e.g. the channel is closed during shutdown), the client gets a 500 but a `queued` row remains. The startup re-queue path (#192) can recover it — but a **client retry in the meantime creates a second task** for the same work. That's the at-least-once / duplicate-submission case this dedup story should cover: idempotency keyed on `(key_id, target_address, transition_index)` so a retry collapses onto the existing task rather than creating a duplicate. Flagging so it's on the radar for the dedup design.

bagelface:
Folded into gas-killer/roadmap#19 as #390.

Worth recording that this is a **correctness** fix as much as a load fix. Once there is at most one in-flight task per target contract (#377), duplicate submissions queue behind each other for the **same transition index**, so the second is guaranteed to be wasted work — and before #378 lands, it renders a payload known to revert with a silent 10M-gas fallback.

Two implementation notes from the analysis:

[...]

PR #302, opened 2026-07-03
feat: add `POST /tasks` async submission returning task IDs

### LINKED ISSUE #193: Async task submission returns task IDs instead of fire-and-forget

bagelface:
## Summary

Switches task submission to an async, persisted model: `POST /tasks` validates a request, writes it to the store as a `queued` task **before** responding, enqueues it for aggregation, and returns `202 Accepted { "task_id": "<uuid-v4>", "status": "queued" }`. `POST /trigger` is kept as a deprecated alias routed to the same handler, so existing clients get the new shape and can migrate.

Closes gas-killer/service#193 (TASK-1). Part of the Service v1 Beta epic (gas-killer/roadmap#15).

> **Stacked on #300.** This branch builds on `bagelface/task-store-model` (the persistent task model) and targets it as its base. Retarget to `main` once #300 merges.

## What changed

- **`POST /tasks`** (new primary endpoint) and **`POST /trigger`** (deprecated alias) both route to `submit_task_handler`, which returns `202` with `{ task_id, status }`.
- Task IDs are UUID v4, generated and persisted by the store's `create_task` (from #300).
- **Persist-before-response:** the task row is written before the `202` is sent, so a restart can recover work already acknowledged to the client (satisfies the persistence ordering criterion of #192).
- **Order preserved:** auth → load-shed (`503` queue full) → validation (`400`/`422`) → onchain validation → persist → enqueue. All rejections stay synchronous and run before any row is written.
- `authorize_task_request` now returns the authenticated key's id, so the persisted task is scoped to its owning key (`key_id`).
- Submission now requires a configured store: the store-less dev path returns `503 NotConfigured` rather than silently accepting and dropping a request.
- Removed the old `{ success, message }` response shape.

## Scope

Submission layer only. The task is created as `queued`; the status lifecycle (`processing`/`ready`/`failed`) is **not** wired here — driving those transitions requires threading the task id through the aggregation channel to the orchestrator, and the terminal `ready` payload depends on the user-executable payload work (gas-killer/service#195). That lifecycle wiring lands with the status-polling endpoints (gas-killer/service#194), which also make the status observable. The channel is intentionally left carrying `GasKillerTaskRequest` unchanged in this PR.

## Testing

- Rewrote the submission happy-path tests to run against an in-memory store with a minted key, asserting `202`, a UUID-v4 `task_id`, `status: "queued"`, that the task is queued, and that it is persisted and scoped to the submitting key.
- Added `/tasks` endpoint tests (accepts a valid key + persists; rejects a missing token when a store is present).
- Existing validation (`400`/`422`), load-shed (`503`), and auth (`401`) tests are unchanged and still pass — they reject before the persist step.
- `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings` (clean), and `cargo test --all-targets --all-features` all pass (121 lib tests).

### DIFF
--- a/helm/gas-killer/templates/ingress.yaml
+++ b/helm/gas-killer/templates/ingress.yaml
@@ -31,7 +31,7 @@ spec:
             routed here and stays reachable only in-cluster. Falls back to the router's public
             routes if publicPaths is unset, never to a catch-all "/".
           */}}
-          {{- range .Values.ingress.publicPaths | default (list "/trigger" "/avs-metadata" "/healthz") }}
+          {{- range .Values.ingress.publicPaths | default (list "/tasks" "/trigger" "/avs-metadata" "/healthz") }}
           - path: {{ . | quote }}
             pathType: Prefix
             backend:
--- a/helm/gas-killer/values.yaml
+++ b/helm/gas-killer/values.yaml
@@ -670,6 +670,8 @@ ingress:
   # This keeps admin behind cluster access in addition to ADMIN_KEY. Add a path here only if it
   # genuinely needs to be internet-facing.
   publicPaths:
+    - /tasks
+    # Deprecated alias of /tasks, kept until clients migrate.
     - /trigger
     - /avs-metadata
     - /healthz
--- a/router/src/ingress.rs
+++ b/router/src/ingress.rs
@@ -1,7 +1,7 @@
 use crate::creator::{TaskQueueDepth, TaskSender};
 use crate::error::{ApiError, ApiErrorBody, ApiErrorEnvelope, ApiJson, ErrorCode};
 use crate::metrics::MetricsCollector;
-use crate::store::{ApiKeyMetadata, CreatedApiKey, SqliteStore};
+use crate::store::{ApiKeyMetadata, CreatedApiKey, SqliteStore, TaskStatus};
 use alloy_primitives::{Address, U256};
 use alloy_provider::Provider;
 use axum::{
@@ -161,14 +161,19 @@ fn check_bearer_auth(headers: &HeaderMap, expected: &str) -> bool {
         .is_some_and(|token| constant_time_eq(token.as_bytes(), expected.as_bytes()))
 }
 
-/// Authorizes a task-submission request. When a durable store is configured — always the case
-/// in production — a valid, unrevoked API key is required. With no store, the request is rejected
-/// unless `allow_unauthenticated` is set (the test/dev bare constructor), so a missing store
-/// fails closed rather than silently opening the endpoint.
-async fn authorize_task_request(state: &IngressState, headers: &HeaderMap) -> Result<(), ApiError> {
+/// Authorizes a task-submission request, returning the authenticated key's id when a store is
+/// configured. When a durable store is present — always the case in production — a valid,
+/// unrevoked API key is required and its id is returned so the submitted task can be scoped to
+/// its owner. With no store, the request is rejected unless `allow_unauthenticated` is set (the
+/// test/dev bare constructor), which yields `None`; a missing store fails closed rather than
+/// silently opening the endpoint.
+async fn authorize_task_request(
+    state: &IngressState,
+    headers: &HeaderMap,
+) -> Result<Option<String>, ApiError> {
     let Some(store) = &state.store else {
         return if state.allow_unauthenticated {
-            Ok(())
+            Ok(None)
         } else {
             Err(ApiError::unauthorized())
         };
@@ -176,7 +181,7 @@ async fn authorize_task_request(state: &IngressState, headers: &HeaderMap) -> Re
 
     if let Some(token) = bearer_token(headers) {
         match store.verify_api_key(token).await {
-            Ok(Some(_)) => return Ok(()),
+            Ok(Some(id)) => return Ok(Some(id)),
             Ok(None) => {}
             Err(e) => {
                 tracing::error!(error = %e, "api key verification failed");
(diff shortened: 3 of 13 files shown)


BreadrichEngels (approved):
Clean submission-layer slice that hits every acceptance criterion in #193, and the ordering is exactly right for the persistence goal. Good to merge into the `bagelface/task-store-model` aggregation branch. Flagging a few cross-cutting things the branch needs before it goes to `main` — none blocking this merge into the branch.

**Meets #193:** `/tasks` primary + `/trigger` deprecated alias → same handler; `202 { task_id, status: "queued" }` (UUID v4, asserted in the test); all validation/load-shed/auth still run and reject synchronously *before* any row is written; `TaskStatus` serializes snake_case so the body matches the spec exactly.

**Ordering is correct.** auth → load-shed → validate → onchain → **persist → enqueue**, with `create_task` writing the `queued` row before both the `sender.send` and the `202`. That's the persist-before-ack durability #192 wants, and the `require_store` (503) + `key_id.ok_or(unauthorized)` guards are sound — with a store present, `authorize_task_request` always yields `Some(id)` or errors 401 first, so the `ok_or` is correctly defensive rather than an unwrap. Nice tests: 202 + UUID-v4 + persisted-and-scoped-to-key, both routes, missing-token 401, and store-less 503.

**Before the branch merges to `main`:**
1. **The response-shape change breaks any client parsing `{ success, message }`.** Most concretely, `scripts/run_scenario.rs` deserializes responses into `struct ApiResponse { success, message }` — a `{ task_id, status }` body will fail that parse and get scored as a failed request. `send_request` and the e2e/helm trigger assertions are in the same boat. Those client tools need updating in lockstep on this branch; otherwise the scenario/e2e suites go red the moment the router returns 202. Worth doing as a sibling PR on the branch.
2. **Sequencing of the follow-ons matters.** Right now every accepted task is persisted `queued` and never transitions (the id isn't threaded to the orchestrator yet — as the scope note says). That's fine in isolation, but the startup re-queue (`incomplete_tasks`) must **not** be wired before the status transitions are, or a restart will re-enqueue and re-run already-completed work. Since #194 lands both together per your plan, just keep them in that order.
3. **Persist-then-enqueue leaves an orphan on send failure.** If `sender.send` fails after `create_task` (channel closed at shutdown), the client gets a 500 but a `queued` row remains — recoverable by the re-queue path, but a client retry then creates a second task. That's the at-least-once/duplicate case the #192 dedup+TTL story (`(key_id, target, transition_index)`) is meant to cover; flagging so it's on the radar for that PR.

**Minor (dev-only):** on the store-less path, a valid request now runs full validation (including the onchain RPC round-trips) before `require_store` returns 503 — you could fail fast by checking the store right after auth. Relatedly, `allow_unauthenticated` is now nearly vestigial since submission requires a store regardless; worth a cleanup pass later.

CI isn't reporting on the head commit yet — confirm `fmt`/`clippy`/the 121 tests are green. Approving for the branch.

bagelface:
Thanks for the thorough review. All four cross-cutting items are now handled — details below.

**1. Client tooling in lockstep → sibling PR #304** (`bagelface/scenario-tasks-endpoint`, based on this branch so it travels to `main` together and can't merge independently).

One correction after digging into each tool: they weren't all in the same boat.
- `scripts/run_scenario.rs` — the real break, as you called out: it deserialized `{ success, message }` and gated on HTTP `200`, so every accepted task would fail to parse and score as failed. Fixed: parses `{ task_id, status }`, treats `202` as accepted, surfaces the error envelope verbatim on non-`202`, submits to `/tasks`.
- `scripts/send_request.rs` — **not** a parse break: it only checks `status.is_success()` (`202` qualifies) and verifies via on-chain `currentSum`; it never deserializes the body. Repointed its default URL to `/tasks` for consistency.
- e2e / helm suites — **not** a break either: both drive `send_request` and parse only `.key` from the admin endpoint, and `run_helm_test.sh` reaches the router via `kubectl port-forward` (bypassing the ingress). Comments updated to `/tasks`.

One thing your list didn't mention that I caught while there: the Helm ingress `publicPaths` allowlisted `/trigger` but not `/tasks`, so the new primary endpoint would be unreachable through the public ingress once this deploys. #304 adds `/tasks` to `publicPaths` (values + template default) and keeps `/trigger` until it's removed.

**2. Follow-on sequencing — agreed and locked in.** The startup re-queue (`incomplete_tasks`) will not be wired before the status transitions exist; #194 lands the transitions and the re-queue together, transitions first, so a restart can't re-run completed work. Recorded as a hard sequencing constraint on my side.

**3. Orphan-on-send-failure / duplicate → flagged on the dedup story (#197).** Posted a note there: persist-then-enqueue can leave a `queued` row if `sender.send` fails, and a client retry then creates a second task — the at-least-once case dedup on `(key_id, target_address, transition_index)` should collapse. https://github.com/gas-killer/service/issues/197#issuecomment-4876940833

**4. Dev-only cleanups → tracked in #306.** Both together, since the store-less validation tests are what pin `require_store` late: migrate those tests to a store-backed harness, then move the store check right after auth to fail fast, and retire the now-vestigial `allow_unauthenticated`.

**CI / green check:** locally `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings` (0 warnings), and `cargo test --all-targets --all-features` all pass — 121 lib tests including the new 202/UUID-v4/persisted-and-scoped and both-routes cases.

BreadrichEngels (approved):
Re-review — the response-shape breakage I flagged is now handled comprehensively in-PR, and you also fixed a public-routing gap I'd have flagged next. The core handler is unchanged from my prior approval, so this is just confirming the follow-through.

**✅ Client tooling migrated in lockstep (my note #1).** Rather than leaving it for a sibling PR, this PR now carries the whole cutover, and it's done right:
- `run_scenario.rs`: `ApiResponse{success,message}` → `AcceptedResponse{task_id,status}`, posts to `/tasks`, `passed()` gates on `202`, and the parse logic cleanly separates the three cases — `202`+parseable → success (message shows `status (task <id>)`), `202`+unparseable → failure, non-`202` → failure surfacing the error envelope verbatim. On-chain verify is re-gated to `202`. The scenario `.toml` doc comments and `send_request.rs`'s default URL all move to `/tasks` too.
- So the scenario/e2e suites won't go red on the 202 — exactly the lockstep I was worried about, closed inside this change.

**✅ `/tasks` added to the Ingress allowlist (nice catch).** `ingress.yaml`'s default list and `values.yaml`'s `publicPaths` both now include `/tasks` (with `/trigger` kept as the deprecated alias "until clients migrate"). Without this, the new primary endpoint would've been in-cluster-only and unreachable by public clients — good that you caught it; both the template default and values are consistent.

The submission handler itself (auth→load-shed→validate→onchain→persist→enqueue, persist-before-ack, defensive `key_id.ok_or`) is byte-identical to what I approved, and the tests still cover 202 + UUID-v4 + persisted-and-scoped, both routes, and store-less 503.

My other two forward-looking notes still stand as **branch-level** items for the follow-on PRs (not this one): don't wire the startup re-queue (`incomplete_tasks`) before the status transitions land, and the persist-then-enqueue send-failure leaves a `queued` orphan that the #192 dedup/TTL story should reconcile. Nothing to change here for those.

Base is now the `task-lifecycle-refactor` aggregation branch, as intended. CI still isn't reporting — confirm `fmt`/`clippy`/tests green. Re-approving.

Issue #303, opened 2026-07-03
Remove deprecated `POST /trigger` endpoint


bagelface:
## Parent Epic
gas-killer/roadmap#15

## Background

gas-killer/service#193 (PR #302) made `POST /tasks` the primary submission endpoint and kept `POST /trigger` as a **deprecated alias** routed to the same handler, returning the same `202 { task_id, status }` shape to ease client migration. Once clients have moved to `/tasks`, the alias should be removed to shed the dual-endpoint surface and the "deprecated" caveats scattered through configs and docs.

This is a **breaking change** — it should land only after clients have migrated and the deprecation window has been communicated (coordinate with the Developer docs task in the epic).

## Prerequisite

- #193 shipped and `/tasks` is the documented endpoint.
- Migration window communicated to clients.

## Cleanup surface

Removing `/trigger` touches more than the router — every reference below should move to `/tasks`:

- **Router** — `router/src/ingress.rs`: drop the `/trigger` route (keep `/tasks`) and the `/trigger`-specific tests/helpers (e.g. `bearer_request`/`json_request` default to `/trigger`).
- **Client tooling** — `scripts/send_request.rs`, `scripts/run_scenario.rs`: point the submission URL at `/tasks`.
- **Scenario configs** — `scripts/scenarios/*.toml` (example, local_profile, latency_profile, testnet_load, testnet_smoke, testnet_parallel_smoke): update endpoint references/comments.
- **Test scripts** — `scripts/run_e2e_test.sh`, `scripts/run_helm_test.sh`: submit against `/tasks`.
- **Helm** — `helm/gas-killer/templates/ingress.yaml` `publicPaths` default list and `helm/gas-killer/values.yaml`: replace `/trigger` with `/tasks`.
- **Docs** — `README.md`, `helm/gas-killer/README.md`, `docker-compose.yml`, `example.env`: update curl examples and public-path references.

## Acceptance Criteria

- [ ] `POST /trigger` route removed; only `POST /tasks` accepts submissions (a `/trigger` request returns 404/405)
- [ ] Router tests updated to exercise `/tasks`
- [ ] All client tooling, scenario configs, and e2e/helm test scripts submit to `/tasks`
- [ ] Helm `publicPaths` default and docs reference `/tasks`, not `/trigger`

Issue #301, opened 2026-07-03
Page the startup re-queue backlog load (incomplete_tasks)


bagelface:
## Parent Epic
gas-killer/roadmap#15

## Background

The persistent task model PR (#300) added `SqliteStore::incomplete_tasks()`, which loads **every** `queued`/`processing` task in a single unbounded query:

```sql
SELECT ... FROM tasks WHERE status IN ('queued', 'processing') ORDER BY created_at, id
```

This is the query the startup re-queue path (an acceptance criterion of #192) will call to rebuild the in-flight backlog after a restart. It is fine at beta volume, but a large backlog would be materialized entirely in memory in one shot.

## Proposal

Page the backlog load so the startup re-queue drains it in bounded batches.

**Prefer keyset (cursor) paging over `LIMIT`/`OFFSET`.** The re-queue loop transitions tasks `queued → processing` as it drains, mutating the `WHERE status IN (...)` set underneath the cursor. Under `OFFSET`, that causes skipped or double-counted rows, and `OFFSET` is O(n) per page. A keyset cursor on `(created_at, id)` — e.g. `incomplete_tasks_after(cursor, limit)` — is stable under that mutation and O(log n) per page.

## Scope / relationship

This is the paging detail of the startup re-queue wiring tracked in #192 (RELY-1). Best done alongside or just before that wiring, since the consumer defines the batch size and drain contract. Not required for correctness at beta volume — this is post-beta scale hardening. Stems from review of #300 (see gas-killer/service#192 comment).

## Acceptance Criteria

- [ ] `incomplete_tasks` (or a keyset successor) returns bounded batches rather than the full set
- [ ] The startup re-queue path drains the backlog in batches until exhausted
- [ ] Test covering a multi-batch drain (backlog larger than one page)

bagelface:
Folded into gas-killer/roadmap#18 #380.

The backlog gets larger with concurrency: up to `MAX_QUEUE_DEPTH` queued rows plus W in-flight `processing` rows, all loaded at once by `incomplete_tasks()` before the sequencer starts.

#380 pages the load, and adds two checks that should run as rows are paged rather than after: reject anything already past its deadline (settle it `expired` instead of spending an EVMSketch on it), and skip anything whose assigned height already certified (#325).

Issue #196, opened 2026-05-05
Task expiry and TTL management


bagelface:
## Parent Epic
gas-killer/roadmap#15

## Background

Tasks include a `block_height` reference that becomes stale as new blocks are produced. A task sitting in queue for too long will fail on-chain validation. An explicit TTL catches this before wasting an aggregation round on a doomed task.

## Acceptance Criteria

- [ ] Configurable `TASK_TTL_SECONDS` env var (default: 300s / 5 minutes)
- [ ] Tasks not moved to `processing` within the TTL are moved to `expired` with reason `QUEUE_TTL_EXCEEDED`
- [ ] `ready` tasks expire after the same TTL from the time they became ready — the payload's `valid_until_block` is the authoritative on-chain deadline, but this cleans up server state
- [ ] A background task sweeps for expired entries every 60 seconds
- [ ] Expired tasks are counted in a Prometheus counter: `gas_killer_tasks_expired_total`

## Metadata

| Field | Value |
|-------|-------|
| Epic | Task Lifecycle Redesign |
| Story ID | TASK-4 |
| Complexity | Low |
| Priority | P1 — strong recommendation |

bagelface:
Follow-on note from the task model PR (#300): the persistent `Task` model adds `TaskStatus::Expired` as a terminal state whose schema comment and enum doc both assert it "carries an error" — but there is no `mark_task_expired(id, error)` setter analogous to `mark_task_failed`. The only current path to `Expired` is `update_task_status`, which sets status only and cannot record an error. When wiring the TTL sweep here, add `mark_task_expired(id, error)` (mirroring `mark_task_failed`) so the expired-carries-error contract is actually expressible.

bagelface:
Folded into gas-killer/roadmap#19 as #390, which wires this into the scheduler rather than treating it as a standalone sweeper.

Two refinements from the analysis:

- The sweeper's job is specifically the case admission control **cannot** catch: "we said yes, then the system got slower." It must remove the task from the scheduler and **release its lane slot before dispatch**, so the freed capacity goes to work that can still succeed. Spending an EVMSketch on a task that is already past its deadline is the concrete waste.
- Do **not** expire `processing` tasks mid-round. The aggregation height is already assigned and must resolve either way, so cancelling the row while the round proceeds only loses the attribution. Let it settle, then record the expiry.

[...]

PR #305, opened 2026-07-03
Task lifecycle refactor

bagelface:
## Summary

Implements the async task lifecycle for v1 beta (gas-killer/roadmap#15): a persistent task model, async submission, status polling, and lifecycle wiring with restart recovery. Covers gas-killer/service#192, #193, #194, and #308 — all already closed individually as their pieces landed on this branch, ahead of this stack reaching `main`.

## What changed

- **Persistent task store (#192)** — `router/migrations/0002_tasks.sql`, a `Task`/`TaskStatus` model, and `SqliteStore` CRUD (`create_task`, `get_task`, `list_tasks_for_key`, `mark_task_ready`, `mark_task_failed`, `update_task_status`, `incomplete_tasks`). Write-before-ack: a task is persisted and has an id before `POST /tasks` responds. Store path is `DATA_DIR`-configurable (default `/data`), PVC-backed in Helm.
- **`POST /tasks` (#193)** — submission returns `202 {task_id, status: "queued"}`. Requires an authenticated key and a configured store (`503` without one). `/trigger` is kept as a deprecated alias routing to the same handler; client tooling (`run_scenario.rs`, scenario TOMLs) and the ingress `publicPaths` allowlist were migrated to `/tasks` alongside it.
- **`GET /tasks/{id}` + `GET /tasks` (#194)** — full task state (status, timestamps, `error`, `payload`), `404` for unknown ids, `403` for a task owned by a different key. Listing is scoped to the caller's key, newest first, with an `?status=` filter and clamped `limit`/`offset` pagination. Added an `ApiQuery` extractor and a `Forbidden` error code so query-string and ownership failures share the same error envelope as the rest of the API.
- **Lifecycle wiring + restart recovery (#308)** — `sequencer.rs`/`executor.rs` now drive `queued → processing → ready/failed` through the `commonware-consensus` aggregation engine, via a single `InFlightTask` slot shared between the task source and the verification handler (valid because the engine's `Sequencer` drives exactly one height at a time). On startup, `factories::requeue_incomplete_tasks` re-enqueues anything left `queued`/`processing` by a previous router life before the sequencer starts pulling.
- **e2e** — `scripts/send_request.rs` now polls `GET /tasks/{id}` until it reaches `ready` (or fails/times out) instead of only checking that the on-chain value changed, so the status-transition path is actually asserted in CI.

## Scope / follow-ups

- `payload` stays `null` until gas-killer/service#195 (executable payload) lands.
- `expired` is untouched — TTL/sweep is tracked separately in gas-killer/service#196.
- A known, accepted at-least-once edge on the restart re-queue (a task whose on-chain transition already landed can be re-run and misreported `failed`) is documented in `factories.rs` and tracked in gas-killer/service#325.
- Deliberately deferred as their own issues: dedup on retried submissions (#197), keyset pagination for `incomplete_tasks` (#301), removing the `/trigger` alias (#303), ingress auth cleanup (#306).

## Testing

`cargo fmt --all`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test --all-targets --all-features` all pass on this branch.


BreadrichEngels (approved):
This is the roll-up of the async task lifecycle to `main`, and I reviewed each slice in depth as it landed on the branch — the store model (#300), `POST /tasks` submission (#302), the `GET /tasks` polling endpoints (#307), and the lifecycle wiring + startup re-queue (#324). Confirming the aggregate here.

**The aggregate is a clean union of the already-approved pieces — no scope creep.** All 26 files map to one of those slices: `store/tasks.rs` + `0002_tasks.sql` + `store/mod.rs` (#300); `ingress.rs` + `error.rs` (the `POST /tasks` handler, the `GET /tasks`/`GET /tasks/{id}` handlers, `ApiQuery`, the `Forbidden` code — #302/#307); `sequencer.rs` + `executor.rs` + `factories.rs` + `main.rs` (the `InFlightTask` slot, the `queued→processing→ready/failed` wiring, `requeue_incomplete_tasks` — #324); and the client-tooling migration to `/tasks` (`send_request.rs`, `run_scenario.rs`, the scenario TOMLs, the e2e/helm scripts, the Helm `publicPaths` allowlist, the `uuid` dep). No engine-migration files snuck in — that's already on `main`.

**What I verified across the slices holds together:** write-before-ack (task persisted with an id before the 202); every task path settles (success→`ready`, enrich/validate/execution failure→`failed`, quorum-skip→orphan-settled `failed`) with no double-settle, on the single-slot model justified by the engine's one-height-at-a-time `Sequencer`; the read endpoints are key-scoped with clamped pagination and `404`/`403`/`401`; and — importantly — the re-queue lands *with* the transitions (transitions-first), so #192's restart-recovery criterion is met without re-running completed work blind.

**Good that the e2e actually exercises the lifecycle now.** `send_request.rs` polling `GET /tasks/{id}` to `ready` (rather than only checking the on-chain value) means the status-transition path is asserted in CI end-to-end — that's the real integration gate for a to-`main` merge, so make sure that run is green.

**The cumulative cutover for `main` (call it out in the release notes):** this changes the submission contract wholesale — `POST /tasks` returns `202 {task_id, status}`, `/trigger` is aliased to the new shape, and submission now *requires* a configured store (PVC-backed) plus an authenticated key. Every in-repo client (send_request, run_scenario, scenarios, e2e, ingress allowlist) is migrated in-branch so the repo is internally consistent, but any **external** caller still parsing `{success, message}` or expecting a fire-and-forget `/trigger` breaks. Combined with the API-key auth already on `main`, a client that isn't re-keyed *and* updated to the async shape is doubly broken — worth a heads-up to any live testnet consumers before this deploys.

**Follow-ups are tracked, and they capture the review feedback cleanly** — #325 for the at-least-once re-run edge (a task whose on-chain transition already landed can be re-run and misreported `failed`; the contract's transition-index idempotency prevents double-apply, so it's a status-accuracy wart, not a safety bug), #301 for the unbounded `incomplete_tasks` scan I flagged on #300, plus #195 (payload, still `null`), #196 (`expired`/TTL), #197 (submission dedup), #303 (drop the `/trigger` alias), #306 (auth cleanup). One thing to keep honest as this evolves: the single-slot attribution rests on the `Sequencer` never pipelining two heights — worth a guard-comment there so a future engine change can't silently break task settlement.

I'm approving on the strength of the per-slice reviews plus this coherence check; I did not re-read all 116KB line-by-line. Get the full workspace + the (now lifecycle-asserting) e2e green before merging to `main` — CI isn't reporting on the head commit yet. Nice, disciplined landing of a big feature.

PR #307, opened 2026-07-03
feat: add task status polling endpoints (GET /tasks, GET /tasks/{id})

### LINKED ISSUE #194: Task status polling endpoint

bagelface:
## Summary

Adds the status-polling API from #194 (TASK-2): `GET /tasks/{task_id}` for a single task and `GET /tasks` for the calling key's list, both scoped to the authenticated API key.

Closes gas-killer/service#194. Part of the Service v1 Beta epic (gas-killer/roadmap#15). Targets the `task-lifecycle-refactor` aggregation branch (built on #300/#302/#304).

## Endpoints

- **`GET /tasks/{task_id}`** → full task state; `404` for an unknown id, `403` for a task owned by a different key, `401` without a valid key.
- **`GET /tasks`** → the calling key's tasks, newest first; `?status=` filter and `?limit=`/`?offset=` pagination (`limit` defaults to 50, clamped to 200; negative `offset` treated as 0).
- Response shape: `{ task_id, status, created_at, updated_at, error, payload }` — `payload` populated when `ready`, `error` when `failed`/`expired`.

Both read straight through the store methods added in #300 (`get_task`, `list_tasks_for_key`). Supporting bits: a new `Forbidden` error code (403) + `ApiQuery` extractor (so a bad `?status=` returns the JSON error envelope, not axum's plaintext 400), and the auth helper renamed `authorize_task_request` → `authenticate_caller` since it now serves both submit and read paths.

## Two decisions worth a look

1. **Timestamps as unix seconds (`i64`), not RFC3339.** The #194 spec example shows `"2026-04-27T12:00:00Z"`, but the rest of the API already returns unix seconds (the admin key listing's `created_at`/`last_used`/`invalid_at`), and there's no date crate in the workspace. I went with `i64` for consistency and to avoid adding a dependency. Easy to switch to RFC3339 if you'd rather match the example — say the word.
2. **`403` vs `404` for a task you don't own.** Per the spec, an unknown id is `404` and a task owned by another key is `403`. That does reveal that *some* task exists at that id to a non-owner. If you'd prefer not to leak existence, I can collapse non-owner to `404`.

## Scope — read endpoints only

The status lifecycle is **not** wired here: accepted tasks are persisted `queued` and no code transitions them yet, so polling reflects whatever is in the store (today, `queued`). The handlers fully support every state — tests drive `ready`/`failed` by transitioning tasks directly through the store and assert the payload/error surface correctly — so #194's contract is met.

Making transitions actually happen (thread the `task_id` through the aggregation channel → orchestrator drives `processing`/`ready`/`failed`) plus the startup re-queue is the next PR, tracked separately. **Hard ordering constraint carried over:** the re-queue (`incomplete_tasks`) must not be wired before those transitions, or a restart re-runs completed work.

## Testing

- 11 new integration tests through the real axum router: owner fetch, `ready`→payload, `failed`→error, `404`, `403` (other key), `401`; list scoping + newest-first, `?status=` filter, pagination, bad-status→enveloped `400`, `401`.
- `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings` (0 warnings), `cargo test --all-targets --all-features` all pass (132 lib tests).

### DIFF
--- a/router/src/error.rs
+++ b/router/src/error.rs
@@ -11,9 +11,10 @@
 //! any time. Status codes are carried alongside the code internally and are unchanged
 //! from the per-endpoint behaviour they replace.
 
-use axum::extract::rejection::JsonRejection;
-use axum::extract::{FromRequest, Request};
+use axum::extract::rejection::{JsonRejection, QueryRejection};
+use axum::extract::{FromRequest, FromRequestParts, Query, Request};
 use axum::http::StatusCode;
+use axum::http::request::Parts;
 use axum::response::{IntoResponse, Response};
 use axum::{Json, async_trait};
 use serde::{Deserialize, Serialize};
@@ -41,6 +42,9 @@ pub enum ErrorCode {
     QueueFull,
     /// The request is missing valid authentication credentials.
     Unauthorized,
+    /// The caller is authenticated but not permitted to access the requested resource (e.g. a
+    /// task owned by a different API key).
+    Forbidden,
     /// An upstream RPC endpoint failed or is unreachable.
     RpcUnavailable,
     /// No contract is deployed at the requested target address on any supported chain.
@@ -101,6 +105,16 @@ impl ApiError {
         )
     }
 
+    /// 403 with [`ErrorCode::Forbidden`].
+    pub fn forbidden(message: impl Into<String>) -> Self {
+        Self::new(StatusCode::FORBIDDEN, ErrorCode::Forbidden, message)
+    }
+
+    /// 404 with [`ErrorCode::NotFound`].
+    pub fn not_found(message: impl Into<String>) -> Self {
+        Self::new(StatusCode::NOT_FOUND, ErrorCode::NotFound, message)
+    }
+
     /// 503 with [`ErrorCode::QueueFull`].
     pub fn queue_full(message: impl Into<String>) -> Self {
         Self::new(
@@ -158,6 +172,32 @@ where
     }
 }
 
+/// A `Query<T>` extractor that emits the [`ApiErrorEnvelope`] on failure instead of axum's
+/// default plain-text body, so a malformed query string (an unknown `?status=` value, a
+/// non-integer `?limit=`) shares the same error contract as the rest of the API. The status
+/// from the underlying [`QueryRejection`] is preserved (400).
+pub struct ApiQuery<T>(pub T);
+
+#[async_trait]
+impl<S, T> FromRequestParts<S> for ApiQuery<T>
+where
+    Query<T>: FromRequestParts<S, Rejection = QueryRejection>,
+    S: Send + Sync,
+{
+    type Rejection = ApiError;
+
+    async fn from_request_parts(parts: &mut Parts, state: &S) -> Result<Self, Self::Rejection> {
+        match Query::<T>::from_request_parts(parts, state).await {
+            Ok(Query(value)) => Ok(ApiQuery(value)),
+            Err(rejection) => Err(ApiError::new(
+                rejection.status(),
+                ErrorCode::InvalidRequest,
+                rejection.body_text(),
+            )),
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -173,6 +213,7 @@ mod tests {
             (ErrorCode::RateLimited, "RATE_LIMITED"),
             (ErrorCode::QueueFull, "QUEUE_FULL"),
             (ErrorCode::Unauthorized, "UNAUTHORIZED"),
+            (ErrorCode::Forbidden, "FORBIDDEN"),
             (ErrorCode::RpcUnavailable, "RPC_UNAVAILABLE"),
             (ErrorCode::ContractNotFound, "CONTRACT_NOT_FOUND"),
             (ErrorCode::InvalidRequest, "INVALID_REQUEST"),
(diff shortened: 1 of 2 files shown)


BreadrichEngels (approved):
Clean, well-tested status-polling slice — meets every acceptance criterion in #194, and the key-scoping is exactly right. LGTM. Answers to your two flagged decisions below.

**Meets #194:** `GET /tasks/{id}` returns the full `TaskView`; `payload` surfaces on `ready` and `error` on `failed`/`expired` (both tested via direct store transitions); `GET /tasks` is key-scoped, newest-first, with `?status=` filter and `?limit`/`?offset`; `404` unknown / `403` non-owner / `401` no key. The `ApiQuery` extractor mirrors `ApiJson` so a bad `?status=` returns the JSON envelope instead of axum's plaintext — nice consistency, and `FromRequestParts` (not `FromRequest`) is the right trait for a query extractor.

**Security holds up:** both handlers authenticate → `require_store` → `key_id.ok_or(unauthorized)` (the same sound defensive pattern as submit), the single-task path enforces ownership before returning, and the list path is scoped to `key_id` in the store query so one key can't see another's tasks. Good touch: `limit` is clamped to `[1, 200]` (default 50) and `offset.max(0)`, so a `limit=1000000` or negative offset can't turn the listing into an unbounded scan. The `/tasks` prefix in the ingress allowlist (from #302) already covers `/tasks/{id}`, so no ingress change is needed here — correct.

**Your two decisions:**
1. **403 vs 404 for a non-owned task — keep the 403.** It's exactly what #194's AC specifies ("403 if the requesting key does not own the task"), so collapsing to 404 would actually deviate from the spec. The existence leak is real but negligible here: task ids are UUID v4 (122 bits of entropy, unenumerable), so a caller can only learn "exists but not yours" about an id they already possess. Not worth trading spec-compliance for. Good that you surfaced it, but I'd leave it.
2. **i64 unix seconds vs RFC3339 — I'd keep i64 and update the spec example.** The #194 example shows an ISO string, but every other timestamp the API emits (the admin key listing's `created_at`/`last_used`/`invalid_at`) is unix seconds, and pulling in a date crate purely to format two fields is a dependency for cosmetics. Internal consistency wins for me; the example was illustrative. Only switch if the *public* client contract (docs/SDK) commits to RFC3339 — your call, but I'd sync the issue's response-shape to `i64` rather than the code to RFC3339.

**Minor (non-blocking):**
- No test exercises the `limit` clamp bounds (`limit>200 → 200`, `limit<1 → 1`) or negative offset — the pagination test uses `limit=2`. A quick assertion on the clamp would lock that behavior in.
- The `expired`-carries-`error` criterion is satisfied structurally (same `error` field as `failed`), but it's untested and not yet reachable — there's still no `mark_task_expired(id, error)` in the store (flagged on #300). Worth pairing when the expiry path lands.

Scope discipline is good as before — read-only, lifecycle unwired, and you've re-carried the hard ordering constraint (don't wire the `incomplete_tasks` re-queue before the status transitions, or a restart re-runs completed work). Targets the `task-lifecycle-refactor` branch as intended. CI isn't reporting on the head commit yet — confirm `fmt`/`clippy`/the 132 tests are green. Approving.

bagelface:
Thanks — both handled.

**1. Timestamps: kept `i64` unix seconds, synced the spec.** Agreed on internal consistency over a cosmetic date dependency. Rather than change the code, I updated #194's response-shape example to unix seconds and added a note that the ISO-8601 example was illustrative. If the public client contract (docs/SDK) later commits to RFC3339, that's the point to revisit — but the API stays uniform for now.

**2. Pagination clamp bounds — now tested.** Extracted the inline clamps into `clamp_page_limit`/`clamp_offset` and added unit tests locking in the bounds: `> MAX_TASK_PAGE_SIZE → 200`, `0`/negative `→ 1`, in-range and `== MAX` pass through, `None → 50`; and offset `None`/negative `→ 0`, passthrough otherwise. Pushed to the branch (`74a2f60`).

Local gate green: `fmt`, `clippy --all-targets --all-features -- -D warnings` (0 warnings), `cargo test` — 135 lib tests.

PR #309, opened 2026-07-03
feat: ECDSA signature aggregation via EigenLayer's ECDSAStakeRegistry

RonTuretzky:
Built on #299 (commonware-consensus aggregation engine migration, included here), this PR replaces BN254 BLS signature aggregation with secp256k1 ECDSA end to end: nodes sign task digests with their registered EigenLayer operator ECDSA keys, and certificates carry a signer bitmap plus per-signer 65-byte `r||s||v` signatures (new `common/src/ecdsa` commonware `certificate::Scheme`, keyed by operator address for both p2p identity and participant ordering).

On-chain verification now delegates to EigenLayer's `ECDSAStakeRegistry` (eigenlayer-middleware v1.5.0, vendored as a git submodule under `contracts/`): `GasKillerSDK.verifyAndUpdate` forwards the quorum to the registry's ERC-1271 `isValidSignature`, which checks each signature against the operator's registered signing key at the reference block and enforces a stake-weight threshold, replacing the entire `BLSApkRegistry`/`NonSignerStakesAndSignature` submission path. The new `contracts/` Foundry project also adds a concrete `GasKillerServiceManager` (`ECDSAServiceManagerBase`) wired to the `AVSDirectory`, plus forge scripts (`DeployECDSAStack` / `RegisterOperatorECDSA` / `FinalizeECDSAStack`) that the EigenLayer setup image runs: it deploys the registry + service manager, registers operators with real AVSDirectory signatures, and sets the threshold to 66% of total weight — so `avs_deploy.json` carries `ecdsaStakeRegistry` from setup and `deploy_array_summation` only reads that address and deploys the ArraySummation target (see the companion PR BreadchainCoop/eigenlayer-bls-local#38).

Configs, docker-compose, helm, and the e2e scripts switch to the operators' `*.private.ecdsa.key.json` files and an address-based `public_orchestrator.json`. Validation: 187 Rust tests, `clippy -D warnings`, `cargo fmt --check`, and 17 forge tests — including a Rust↔Solidity parity anchor proving a Rust-produced signature verifies through the real `ECDSAStakeRegistry` path.


RonTuretzky:
✅ **Full docker-compose e2e validated on-chain** (`run_e2e_test.sh`, router ingress remapped to 18080):

- `ECDSAStakeRegistry` + `GasKillerServiceManager` deployed on the Sepolia fork; all 3 operators registered through the **real AVSDirectory** (Rust-signed EIP-712 registration digests validated on-chain); stake threshold set to 66% of total weight.
- Message-hash parity verified against the deployed contract.
- Live task: nodes ECDSA-signed the digest, the aggregation engine assembled a 3-of-3 certificate, and `verifyAndUpdate` verified it through the real `ECDSAStakeRegistry.isValidSignature` (ecrecover per signature + signing-key checkpoints at the reference block + 300e12 signed weight ≥ 198e12 threshold) — **`currentSum` 0 → 1352, e2e exit 0**.

The e2e's pass condition polls `currentSum` on-chain, which only mutates after the registry call succeeds — so on-chain ECDSA verification is load-bearing for the test.

RonTuretzky:
## Update: ECDSA AVS bootstrap moved into the setup image

Addressing review feedback that the earlier approach (deploying the ECDSA stack from `deploy_array_summation.rs`) broke the "setup container owns all AVS infra" pattern. The stack deployment + operator registration now live in the eigenlayer setup image, so `avs_deploy.json` carries `ecdsaStakeRegistry` + `gasKillerServiceManager` from the start and `deploy_array_summation` is back to its original thin shape (read registry address, deploy target).

**Companion PR:** BreadchainCoop/eigenlayer-bls-local#38 — deploys `DeployECDSAStack` / `RegisterOperatorECDSA` / `FinalizeECDSAStack` (new `contracts/script/`) during setup. This PR's E2E and Helm CI are pinned to that PR's image build (`ghcr.io/breadchaincoop/eigenlayer:pr-ron-gas-killer-ecdsa-stack`).

**Merge order:** merge eigenlayer-bls-local#38 first → `:latest` republished with the ECDSA stack → repin this PR's CI from the `pr-` tag to `:latest` → merge here.

**Validation:** full docker-compose e2e green against a local build of the companion image — the setup container deploys the ECDSA stack + registers 3 operators via the real AVSDirectory + finalizes the threshold, and a live task's ECDSA quorum certificate executes on-chain via `ECDSAStakeRegistry.isValidSignature` (`currentSum` 0 → 1352, exit 0). The prior CI failures (E2E/Helm) were pulling the BLS-era `:latest`; they'll go green once #38's image publishes and this branch's CI re-runs.

RonTuretzky:
## Addressed Copilot's review

**Fixed:**
- Removed a stray nested `contracts/contracts/` directory and a committed `script/deployments/ecdsa-stack/100.json` deploy artifact (a `cd` slip during debugging placed the `.gitkeep` one level too deep). The `.gitkeep` now sits at `contracts/script/deployments/ecdsa-stack/` and `contracts/.gitignore` ignores `broadcast/` + deployment JSONs. (Not in Copilot's comments, but its file list surfaced the bad paths — thanks.)
- **config.rs:370** — corrected the stale comment; the ECDSAStakeRegistry-backed `GasKillerSDK.verifyAndUpdate` **does** enforce `FutureBlockNumber`/`StaleBlockNumber` on-chain (the comment predated switching back to the middleware SDK).
- **StateChangeHandlerLib.sol:55** — `TOOD` → `TODO`.

**Not changed (respectfully disagree):**
- The 5 `require(condition, CustomError())` comments (GasKillerSDK.sol:70/94, StateChangeHandlerLib.sol:42/106/113): this **is** valid Solidity as of **0.8.26** (`require(bool, error)`), and the project pins **solc 0.8.29**. It compiles and all 17 forge tests pass in CI — so these are false positives from a pre-0.8.26 knowledge cutoff.
- **eigenlayer.rs:133** (accept `gasKillerServiceManager` as a service-manager alias): that address feeds the `QUORUM_THRESHOLD()` / `THRESHOLD_DENOMINATOR()` reads, which only `avsServiceManagerWrapper` implements. `gasKillerServiceManager` (`ECDSAServiceManagerBase`) has no such functions, so aliasing it there would revert operator discovery at startup. `avs_deploy.json` still carries `avsServiceManagerWrapper` (the setup image keeps the BLS-middleware registry for operator discovery + p2p sockets), and green e2e/Helm CI confirms discovery works.

RonTuretzky:
## Addressed a deeper adversarial review (beyond Copilot)

Ran a 4-lens migration-completeness review with each finding adversarially re-verified against the code. Confirmed + fixed in `c53f01e`:

- **🔴 Critical — fail-open threshold window.** `DeployECDSAStack` initialized the `ECDSAStakeRegistry` threshold at **0** and `FinalizeECDSAStack` raised it later. The registry reads the threshold *at the caller's reference block*, and `verifyAndUpdate` accepts reference blocks up to `blockStaleMeasure` (~300) old — so for ~300 blocks after finalize, a reference block in the `[firstRegistration, finalize)` window would let a **single** operator satisfy the quorum. Now initialized **fail-closed** (`type(uint224).max`); a new forge test (`test_bootstrapThresholdWindowIsFailClosed`) proves a full quorum is rejected pre-finalize and accepted post-finalize (18 forge tests green).
- **Helm installs broken off-CI.** `eigenlayer.contracts.avsDirectory` became mandatory (setup-job) but was only added to `testnet-overrides.yaml`; added it to `local-overrides.yaml` (so `run_helm_test.sh` works) and `mainnet-overrides.yaml`.
- **deploy_array_summation** now treats `avsServiceManagerWrapper` as optional (prefers `gasKillerServiceManager`), so an ECDSA-only `avs_deploy.json` parses.
- **Migration-window image defaults**: local docker-compose + `local-overrides.yaml` default to the `pr-` ECDSA build until #38 lands in `:latest` (CI already pinned; TODOs to revert).
- **README** corrected: the deploy-time bootstrap moved into the setup image; documented the count-quorum vs stake-weight-threshold assumption.

**Corrected merge order (service first):** #309 merges first (its CI is pinned to the `pr-` image, independent of `:latest`) → bump `GAS_KILLER_SERVICE_REF` to service `main` SHA → merge BreadchainCoop/eigenlayer-bls-local#38 (republishes `:latest`) → follow-up PR reverts the `pr-` pins to `:latest`. (Infra-first would break every other consumer of the mutable `:latest`.)

**Documented as known limitations (not fixed here — pre-existing design / hardening follow-ups):**
- *Count-quorum vs stake-weight threshold*: the engine forms a certificate at `n−⌊(n−1)/3⌋` signers; the registry enforces a stake-weight threshold. These agree under roughly-even stake (the deployment assumption); with highly uneven stake a count-quorum certificate can be rejected on-chain — the height is released as a **failed execution** (logged/metric'd, not silently lost). Stake-aware quorum is a follow-up.
- *Discovery vs verification registries*: the participant set comes from the BLS-middleware registry coordinator (operator discovery + p2p sockets) while verification uses the `ECDSAStakeRegistry`. `main.sh` registers every operator in both, so they stay consistent; a startup cross-check is a recommended hardening.

bagelface:
since we're not moving forward with ECDSA signatures, this can be closed

PR #312, opened 2026-07-06
feat: aggregate Schnorr signature scheme (constant-gas quorum verification)

RonTuretzky:
# feat: aggregate Schnorr signature scheme (constant-gas quorum verification)

The Schnorr equivalent of the ECDSA migration. Where the ECDSA path
(**#299** commonware-consensus engine → **#309** ECDSA aggregation via `ECDSAStakeRegistry`,
mirrored on-chain by solidity-sdk **#49**) submits *N* individually-recovered signatures and
pays per-operator stake lookups (**#311** measured ~65k gas for 3 ops, ~30k cached, and it
grows with the operator set), this submits a **single aggregate Schnorr signature** verified
on-chain in **constant gas** — one `ecrecover`, no secp256k1 scalar-mul in Solidity.

This is now a **full live integration**, not just the crypto core: the node and router
binaries run the interactive two-round protocol end-to-end, switchable with
`SIGNATURE_SCHEME=ecdsa|schnorr`, and both the docker-compose e2e and the Helm integration
test run the whole stack once per scheme.

## Why Schnorr aggregate (not the alternatives)

- **FROST / threshold Schnorr** — one group sig, but *non-attributable* (breaks the
  GasKillerSlasher in solidity-sdk **#46**, which slashes "operator set minus non-signers")
  and *count*-thresholded, not stake-weighted. Rejected.
- **Bitmap Schnorr** (N sigs + bitmap) — attributable and fits the engine, but costs the same
  as ECDSA on-chain. No win. Rejected.
- **Aggregate Schnorr, plain key aggregation + PoP (this PR)** — one signature, constant-gas
  verify, *and* attributable via a non-signer bitmap + on-chain aggregate-key subtraction (the
  EigenLayer `BLSApkRegistry` model, ported to secp256k1). The only option that is both cheap
  and slasher-compatible.

## Measured gas (forge, `test/SchnorrStakeRegistry.t.sol`)

| path | 3/3 signed | marginal / non-signer |
|---|---|---|
| **Schnorr (this PR)** | **6,749** | +4,162 (affine point sub) |
| ECDSA base (#311) | 65,027 | +15,581 / signer |
| ECDSA cached (#311) | 30,825 | +6,700 / signer |

Constant in the number of *signers*; cost grows only with *non-signers* (~0 at healthy
participation).

## What's here

**Crypto core — Rust (`common/src/schnorr/`)**
- `mod.rs` — key/point/signature types, the Scribe (Chronicle/MakerDAO) `ecrecover`-trick
  verification convention, a faithful `ecrecover` mirror proven byte-identical to k256's
  recovery, and proof-of-possession.
- `musig.rs` — MuSig2 two-round aggregate signing (dual-nonce, plain key aggregation), the
  `Coordinator`/`Participant` protocol, and per-signer `verify_partial` attribution.
- `wire.rs` — commonware `Read`/`Write`/`EncodeSize` codec for the four channel-2 messages
  (`NonceRequest` / `NonceCommit` / `SignRequest` / `PartialSig`).
- `examples/schnorr_parity_fixture.rs` — deterministic Rust→Solidity vector generator.

**Live wiring — node + router**
- `node/src/schnorr_participant.rs` — replaces the aggregation engine in schnorr mode. Runs
  the participant state machine over p2p channel 2: per `(height, attempt)` nonce sessions
  (consume-once, in-memory so a restart just becomes a non-signer), signer-set authentication
  (each point must map to a known operator identity address), and a locally-derived digest —
  a node never signs a digest it did not compute itself from its own TaskBook + EVMSketch
  validation.
- `node/src/digest.rs` — the announce-vs-skip + validation-retry logic, factored out of the
  ECDSA automaton so both signing paths vouch for byte-identical digests.
- `router/src/schnorr_coordinator.rs` — replaces the engine + reporter in schnorr mode. Drives
  the two rounds per assigned height with per-signer partial verification and attempt-scoped
  suspect exclusion (a node that commits a nonce but never signs is dropped from the next
  attempt), then emits the same certificate-observation shape the submitter consumes.
- `router/src/schnorr_submitter.rs` + `executor.rs` — submit the single aggregate signature
  `(s, Raddr)` + the ascending non-signer list through `SchnorrGasKillerSDK.verifyAndUpdate`,
  ERC-165-gated on the Schnorr interface id.
- `router/src/sequencer.rs` — generalized over a `CertIndex` trait so it is byte-for-byte
  identical in both modes; the coordinator's mailbox and the ECDSA reporter both implement it.
- The **ECDSA path is genuinely unchanged** — the whole engine/reporter/automaton branch is
  preserved verbatim behind `SignatureScheme::Ecdsa`, and its e2e matrix leg proves it.

**Solidity (`contracts/`)**
- `libraries/{Secp256k1,SchnorrVerify}.sol` — affine point math + the Scribe constant-gas verify.
- `SchnorrStakeRegistry.sol` — PoP-gated registration, aggregate key, non-signer subtraction,
  fail-closed staleness (the "watermark snapshot" model from **#311**).
- `SchnorrGasKillerSDK.sol` — `verifyAndUpdate` against the Schnorr registry; the signed digest
  is unchanged, so the slashing/fraud-proof machinery is scheme-agnostic. Adds a
  `getMessageHash` mirror and an `ISchnorrGasKillerSDK` interface id so the router's ERC-165
  preflight can distinguish a Schnorr target.
- `examples/array-summation/SchnorrArraySummation(+Factory).sol` — the e2e app contract over
  the Schnorr SDK.

**Deploy + harness**
- `scripts/deploy_array_summation.rs` — schnorr branch deploys `SchnorrStakeRegistry`,
  PoP-registers each operator from its existing secp256k1 key, then deploys the target —
  registrations strictly before the target so the registry's `effectiveBlock` watermark is
  satisfied at the first verification.
- `docker-compose.yml`, `run_e2e_test.sh`, `run_helm_test.sh`, the Helm chart/values, and both
  CI workflows thread `SIGNATURE_SCHEME` through the whole stack and **matrix over
  `[ecdsa, schnorr]`**.

## What each layer demonstrates (so the green checks mean something)

- **Rust CI (`cargo test`)** — the in-process MuSig2 protocol (full/subset participation,
  bad-partial attribution, lying-coordinator rejection) and the wire codecs. Crypto + protocol
  logic, in isolation.
- **forge tests** — the on-chain constant-gas verify against a **real Rust-generated aggregate
  signature** (byte-for-byte Rust⇄Solidity parity), plus the SDK/registry control flow.
- **E2E (`schnorr` matrix leg)** — the end-to-end proof: a task flows ingress → sequencer →
  the live two-round protocol across 3 node containers → aggregate signature → on-chain
  `verifyAndUpdate` against the real `SchnorrStakeRegistry`, asserted by `currentSum()` changing
  on-chain. There is no ECDSA fallback in schnorr mode, so a broken live path fails this leg
  rather than passing vacuously.
- **Helm integration (`schnorr` matrix leg)** — the same, on a kind cluster.

> Note: this PR targets a stacked feature branch (`RonTuretzky/pr299-ecdsa-signatures`), so the
> e2e/helm `pull_request` triggers were extended to include that base branch — otherwise the
> only checks gating the PR would be Rust CI + Docker image builds, which never exercise the
> live aggregation. Revert that trigger addition once the stack lands on a mainline branch.

## Rogue-key safety

Plain aggregation is linear (so on-chain non-signer subtraction works) but rogue-key-forgeable
*without* a proof of possession — so registration **requires** a PoP, exactly as
`BLSApkRegistry` does for BLS. See `DESIGN.md`.

## Still out of scope (deliberately, per the ask)

Gas optimizations, the registry-cache contract, slashable-message signing, and multi-call mode
(the ECDSA stack's #311 / #310 / solidity-sdk #48 analogues) are **not** in this PR. Also, nonce
single-use is enforced in-memory (consume-by-value; a restart forgets in-flight nonces and the
node becomes a non-signer) — a persistent "nonce spent" marker for crash-during-round durability
is a follow-up, called out in `musig.rs`. Neither affects the aggregation correctness this PR
demonstrates.

## Adversarial crypto review

Four independent skeptics attacked the crypto (rogue-key, the `ecrecover`-trick algebra,
MuSig2 nonce handling, on-chain subtraction + stake), plus a consolidating judge. **No critical
issues.** Every candidate break was dismissed with the exact code guard that stops it (rogue-key
→ mandatory PoP; `ecrecover` forgery → reduces to ECDLP; `sp==0` → impossible mod prime `n`;
weight underflow / point-sub edge cases → all guarded). The one medium finding — round-2
trusting a coordinator-supplied `(b, e)` — was hardened: `SigningContext` carries the public
aggregate nonces and each signer recomputes `b`/`e` locally and aborts on an inconsistent `R`
(`lying_coordinator_rejected`), and the live participant additionally refuses a *different*
context for a session it already signed (nonce-reuse defense).

Test status: `cargo fmt` / `clippy -D warnings` clean; `cargo test` green; forge test: 39
passed; both CI workflows matrix over `[ecdsa, schnorr]`.


bagelface:
Superseded by #323, which ports this Schnorr work onto `main` (no ECDSA scaffolding, contracts consumed from gas-killer/solidity-sdk#58 rather than vendored). Leaving this open for reference.

bagelface:
Since https://github.com/gas-killer/service/pull/323 has been merge in, I'm closing this.

PR #313, opened 2026-07-07
feat: unbounded-mode e2e — ArraySummation way above the block gas limit

RonTuretzky:
Wires gas-analyzer's `SimProfile::UnboundedV1` (gas-killer/gas-analyzer#166) through the service and proves the unbounded claim end to end: **a tracked function whose direct execution cannot fit in a real Ethereum block lands on-chain as one small `verifyAndUpdate`**.

## Measured result (native e2e, in CI)

With `ARRAY_SUMMATION_ARRAY_SIZE=20000`:

| | gas |
|---|---|
| direct `sum([])` execution | **52,769,831** (1.76× the 30M mainnet block limit — unlandable) |
| gas-killer payload applied on-chain | **~77,746** (704-byte payload, ~679× less) |

## What

- **`GK_SIM_PROFILE`** (`chain` default \| `unbounded-v1`) read by `GasKillerValidator` at construction and passed to `call_to_encoded_state_updates_with_evmsketch_profiled`. It is protocol configuration — the router and every node must agree or their independently derived payloads (and signatures) diverge; unknown values panic rather than silently fork the quorum.
- **Native e2e** `common/tests/unbounded_analysis.rs`: spawns `anvil --disable-block-gas-limit --chain-id 11155111`, deploys a 20k-element ArraySummation (~450M-gas constructor), asserts the direct estimate exceeds 30M, and asserts the unbounded-profile validator produces the O(1) payload applying under the block limit. Soft-skips when anvil is absent.
- **Docker e2e**: `ANVIL_EXTRA_ARGS` on the ethereum service, `GK_SIM_PROFILE` on router+nodes, `run_e2e_test.sh` steps 7a/10b (direct-estimate assertion + verifyAndUpdate gas comparison), `scripts/run_unbounded_e2e_test.sh` wrapper, and an **`e2e-test-unbounded` CI job** running the full stack in unbounded mode.
- **Deps**: gas-analyzer pinned to `ron/unbounded-sim-mode` until gas-killer/gas-analyzer#165 + #166 merge (TODO in Cargo.toml).

## Upstream fix surfaced by this test

The unbounded shape gate initially rejected the honest ArraySummation payload as `TooManyStores`: every `trackState` diff carries a Store to the SDK's fixed StateTracker counter slot alongside the consumer's own write. Fixed in gas-analyzer `5f7479d` — the gate now exempts that well-known fixed slot (`UnboundedShape.tracker_stores`).


bagelface:
Superseded by #353, which re-ports this onto current `main`. Closing as reference-only.

The idea here was never implemented on `main` — what went stale was the ground underneath it. gas-killer/gas-analyzer#166 has since merged (so no branch pin is needed), but it renamed `SimProfile::UnboundedV1` to `Unbounded`, replaced the single-slot shape gate with one that prices the payload's transport and dispatch, and grew a sixth `encoding` parameter on `..._profiled`. On this side, `STATE_ENCODING`/`prestate-net` landed, the `ARRAY_SUMMATION_*` env block was retired in favour of the examples manifest, and the e2e matrix became include-only.

**Carried over:** the sim-profile-as-consensus-config idea (now parsed in `common/src/config.rs` beside `state_encoding()`), `ANVIL_EXTRA_ARGS`, the helm surface, and the step 7a/10b gas assertions — which are the actual proof and had no equivalent on `main`.

**Dropped:** the 216-line `Cargo.lock` churn; `run_unbounded_e2e_test.sh` and its `ARRAY_SUMMATION_ARRAY_SIZE` sizing; the duplicated `e2e-test-unbounded` job (now one matrix leg); and `common/tests/unbounded_analysis.rs` with its pasted creation-bytecode fixture, since #166 carries anvil-backed profile tests and artifacts here are built from example-contracts.

**Changed:** the above-block-limit target is `OnchainLife.step(3)` (~49.7M gas, ≤17-word diff) rather than a 20k-element ArraySummation. It reuses the manifest harness, and `life_step_1` passing under `chain` while `life_step_3` fails gives the leg a built-in negative control.

Thanks @RonTuretzky — the assertion design and the profile-must-match-across-the-fleet framing came straight from here.

PR #314, opened 2026-07-07
feat: Schnorr nonce-commitment registry — non-interactive aggregate signing

rubydusa:
## Summary

Removes the interactive MuSig2 nonce round from the aggregate-Schnorr path. Operators pre-commit Merkle roots of **nonce batches** to a new on-chain `SchnorrNonceRegistry` (read the way nodes read stakes); the full nonce points travel p2p and are verified against the registrations. Signing becomes a deterministic one-shot per height — ack-shaped — so Schnorr mode returns to `commonware_consensus::aggregation` following the #299 architecture, as `SIGNATURE_SCHEME=schnorr-precommit`.

Design doc (review map): `docs/schnorr-nonce-registry.md` — includes an implementation-status table with file pointers.

## Key pieces

- **`contracts/src/SchnorrNonceRegistry.sol`** — append-only contiguous batch commitments, authenticated by a PoP-style key signature binding `(chainid, registry, operator, batchIndex, startSlot, count, root)`; the chain never stores or opens nonce points. Foundry suite signs everything in-Solidity (test-only scalar-mult helper) + a Rust↔Solidity `batchMessage` parity vector.
- **`common/src/schnorr/precommit.rs`** — slot mapping `idx = h·A + a`, domain-separated seed→(k1,k2) derivation (batch seeds derive from the operator key, so a node rebuilds its secrets from its key file + on-chain metadata), Merkle batches, registration message.
- **`common/src/schnorr/scheme.rs`** — `SchnorrScheme: certificate::Scheme`: the ack IS the MuSig2 partial for the deterministic committed set S₀; two-form certificate (`Aggregate` = final constant-size signature at full S₀; `Attested` = attributable partial bundle certifying the signer set) resolves the set-bound-partials vs any-quorum-`assemble` mismatch. Completion round (`sign_completion`/`CompletionCollector`) finalizes `Attested` sets — no coordinator message, the certified bitmap IS the set agreement.
- **Invariant N1** ("one partial per slot, ever"): deterministic nonces kill the old safe-by-amnesia property, so `FileSpendJournal` enforces fsync-write-ahead slot binding with torn-tail recovery; restart cannot re-sign a consumed slot under a different context.
- **Wiring**: node rebuilds batches from chain state (root-checked), runs the engine with the ECDSA arm's knobs + a channel-2 gossip/completion actor; router runs a verifier engine + one combined reporter/CertIndex/gossip/completion actor feeding the existing schnorr submitter. Registry addresses resolve from env or the deployment JSON — `SIGNATURE_SCHEME=schnorr-precommit ./scripts/run_e2e_test.sh` needs no compose changes.

## Verification

- `forge test` 50/50 (incl. gas benchmark + parity anchors), `cargo test --workspace`, `clippy --workspace -D warnings` clean.
- ECDSA and interactive-Schnorr paths semantically untouched.
- **Draft until**: the docker e2e in `schnorr-precommit` mode is green (run in flight).

## Known follow-ups (deliberately out of scope)

- Delete `NonceRequest`/`NonceCommit` + interactive coordinator/participant actors once e2e parity holds.
- Batch auto re-registration at a consumption threshold — until then a deployment is bounded by batch-0 coverage (`SCHNORR_NONCE_BATCH_SLOTS / 4` heights).
- On-chain root pinning for gossiped batches (closes a per-operator equivocation/stale-directory liveness nuisance; documented in `batches.rs`).
- Completion attempts ≥ 2 (shrink-and-retry); today one completion attempt, then the deadline/skip path.


bagelface:
Superseded by **#343**, which re-targets this work onto `main`.

This PR's base (`RonTuretzky/schnorr-aggregate-signatures`) was superseded on `main` by #323, and retargeting turned out to need a re-port rather than a rebase: merging the base into `main` conflicts in 41 files, largely because `main` *deleted* several files this PR edits — `node/src/reporter.rs`, `node/src/{automaton,task_book}.rs` and `router/src/{reporter,submitter}.rs` all moved upstream into `commonware-restaking`.

#343 carries this PR's own delta, adapted:

- Participant identity is BN254 rather than an ECDSA address (`Engine::start` ties it to the p2p identity, which `main` keeps BN254 in every mode), with Schnorr points index-aligned beside it and `nonSigners` sorted explicitly.
- The `CertifiedSummary` tap and `CertificateInspect` trait are gone — upstream already has a scheme-agnostic `NodeReporter::with_certificate_tap`, which service `main` already pins.
- The router's hand-rolled reporter/CertIndex half is gone — upstream's `CertReporter` is scheme-generic and does it.
- The contracts moved to gas-killer/solidity-sdk#73, since contracts no longer live in this repo. The `sol!` view of `operators(...)` was corrected to the hardened 5-field return along the way.

Design intent, the crypto core and invariant N1 are unchanged. Closing in favour of #343 — thanks @rubydusa; the commit history here stays as the reference for the original work.

Issue #316, opened 2026-07-08
Evaluate enabling CORS on the router ingress (vs. the docs-site proxy)


bagelface:
## Context

The docs site's OpenAPI playground (gas-killer/site#14) lets users fire real requests at the router from the browser. A browser request to the router is a **cross-origin** call, so without CORS headers the browser blocks it.

That's currently solved **entirely on the docs side**: the playground posts to a same-origin Next.js proxy (`app/api/proxy`), which forwards to the router server-side (where CORS doesn't apply). The proxy is scoped to the hosts in `openapi.yaml`'s `servers` (testnet + localhost).

This issue is to decide whether the **router** should instead enable CORS so browsers can call it **directly** — and to record *why we might not*.

## Option A — keep the docs-side proxy (status quo)

- Router CORS policy stays closed (default-deny) — no change to who can call it from a browser.
- The allowlist lives in the repo that needs it (the docs site), next to the `servers` list.
- No service change or deploy; no ongoing config to track docs deployment URLs.
- Cost: playground traffic (including any API key / `ADMIN_KEY` a user types in) transits the docs site's Vercel function on the way to the router.

## Option B — enable CORS on the router (`tower-http` `CorsLayer`)

- Browser talks to the router **directly**: lower latency, no relay, and credentials never touch the docs site's infra.
- The "correct" posture *if* we want third-party browser apps (not just our docs) to call the API.

## Why we might NOT want Option B

1. **It broadens the router's exposure by design.** Our API consumers are primarily **contracts and server-side clients**, not browsers. The only browser client today is our own docs playground. Enabling CORS opts the API into cross-origin browser access we don't otherwise need. CORS is not access control — it only relaxes the browser's same-origin protection; the real gate stays the API-key auth + rate limiting.
2. **Allowlist maintenance.** To be safe you allow specific origins (`https://gaskiller.xyz`, Vercel preview URLs, `localhost`) — a list that now must track the docs deployment across prod/preview/local, in a *different* repo than the docs. The lazy alternative, `Access-Control-Allow-Origin: *`, makes the API callable from **any** website's JS.
3. **Admin surface.** `/admin/keys*` is guarded only by the `ADMIN_KEY` shared secret. A blanket `CorsLayer` also makes the admin endpoints invokable from browser JS (still gated by the secret, but you generally don't want the admin surface reachable from arbitrary web pages). Scoping CORS to the non-admin routes is extra work.
[...]

PR #317, opened 2026-07-09
feat: Ethereum JSON-RPC ingress — submit raw signed txs as Gas Killer tasks

RonTuretzky:
Adds an Ethereum JSON-RPC compatible endpoint to the router so anyone can **switch their RPC URL over to Gas Killer** and submit transactions to permissionless functions unchanged — wallets, ethers/viem, `cast`.

## What

- **`POST /rpc` and `POST /rpc/:key`** (Infura-style path key, since browser wallets can't set an `Authorization` header; same `gk_` API keys as `/trigger`).
- **`eth_sendRawTransaction`** is intercepted: the raw signed tx is decoded (legacy/2930/1559; blob & 7702 rejected), the sender is **ecrecovered from the signature** and becomes the task's `from_address`, target/calldata/value become the task, `block_height` pins to head, transition index auto-resolves. Returns the tx hash. The tx is never broadcast — the state diff lands via `verifyAndUpdate`, so the sender pays zero gas (and no receipt exists for the returned hash).
- **msg.sender**: operators simulate the call with the recovered signer as `msg.sender`, and what lands on-chain is that simulation's storage diff — so sender identity is enforced by key possession at the ingress boundary. (The contract itself can't verify attribution — that's protocol future work; v1 targets permissionless functions.)
- **Read methods** (`eth_chainId`, `eth_getTransactionCount`, `eth_gasPrice`, `eth_call`, … any `eth_/net_/web3_` read) proxy verbatim to the upstream chain RPC (`?chain=l1|l2`), with JSON-RPC 2.0 batch support, so wallet flows (nonce, gas estimation, chain detection) just work.

## Resilience

- Duplicate raw-tx suppression: bounded two-state cache — committed duplicates answer the same hash idempotently; a duplicate racing an in-flight first attempt gets geth-style `already known`; failures roll back so retries work. Upstream work is bounded by `RPC_UPSTREAM_TIMEOUT_SECS`.
- Signed chain id is authoritative: mismatches, unserved chains, and ambiguous same-address multi-chain deployments are rejected; `eth_getCode` probe failures are hard retryable errors (closes an ingress/creator chain-divergence window).
- Wallet-recognizable errors (`insufficient funds…`, `already known`), stable codes (`-32001` auth, `-32002` upstream down, `-32005` at capacity), explicit CORS allow-headers (a `*` wildcard doesn't cover `Authorization`).
- Fixes a pre-existing validator digest-cache collision: the cache was keyed `(transition_index, block_height)` only, so two different tasks in the same block at the same transition count collided; the key now covers target/sender/value/calldata.
- Bumps p2p `peer_connection_cooldown` 1s → 3s (crossed-handshake `DecryptionFailed` flap on virtualized Docker hosts).

## Testing

- 41 new unit/integration tests (protocol conformance, decode/recover paths, chain resolution incl. multi-chain ambiguity, dedupe lifecycle, auth, batches, CORS). Full workspace suite + clippy `-D warnings` green.
- New e2e binary `scripts/send_raw_tx_request` (wired into `run_e2e_test.sh` step 10b): drives every read through the router, signs a real EIP-1559 tx, submits it via the router, asserts hash echo, duplicate idempotency, four negative-path error codes, and on-chain `currentSum`/`stateTransitionCount` effects (including no double-execution from the duplicate).
- Endpoint surface verified against a live local stack (reads, batch, negative paths, acceptance, dedupe, conversion → round dispatch). Round completion is left to this PR's CI e2e — local Docker-VM p2p (`DecryptionFailed`/`Recipients::All` snapshot timing) can't complete rounds for the pre-existing `/trigger` path either.
- Adversarially reviewed (multi-agent): 19 candidate findings, 7 confirmed and all fixed in this PR (dedupe false-success race, unbounded reservation on upstream hang, chain-divergence window, CORS wildcard, digest-cache collision, test gaps).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ version = "0.1.0"
 license = "PPL"
 
 [workspace.dependencies]
-alloy = "1.0"
+alloy = { version = "1.0", features = ["consensus", "eips", "k256", "network", "rlp"] }
 alloy-eips = "1.0"
 alloy-network = "1.0"
 alloy-primitives = "1.0"
@@ -54,7 +54,7 @@ prost = "0.13.5"
 rand = "0.9.1"
 reqwest = { version = "0.12", features = ["json", "rustls-tls"] }
 serde = { version = "1.0.219", features = ["derive"] }
-serde_json = "1.0.140"
+serde_json = { version = "1.0.140", features = ["raw_value"] }
 serde_yaml = "0.9.34"
 sqlx = { version = "0.8", default-features = false, features = ["runtime-tokio", "sqlite", "migrate", "macros"] }
 tempfile = "3"
--- a/README.md
+++ b/README.md
@@ -195,6 +195,53 @@ Set `GAS_KILLER_API_KEY` to a minted key when the router requires auth:
 GAS_KILLER_API_KEY=gk_... cargo run -p scripts --bin send_request
 ```
 
+## Ethereum JSON-RPC Ingress
+
+The router also speaks standard Ethereum JSON-RPC at `POST /rpc`, so existing wallets and
+tooling can submit to Gas Killer by **switching their RPC URL over** — no payload changes:
+
+```bash
+# Wallet-style URL with the API key in the path (for clients that cannot set headers):
+#   http://localhost:8080/rpc/gk_...
+# Or programmatic clients with a Bearer header:
+cast send <target> "sum(uint256[])" "[1,2,3]" \
+  --rpc-url http://localhost:8080/rpc/gk_... \
+  --private-key $PRIVATE_KEY
+```
+
+How it works:
+
+- **`eth_sendRawTransaction`** is intercepted: the signed transaction is decoded, the sender is
+  **recovered from its ECDSA signature** and used as the task's `from_address` (so acting as an
+  address requires holding its key — operators simulate the call with it as `msg.sender`), the
+  target/calldata/value become the task, `block_height` is pinned to the current head, and the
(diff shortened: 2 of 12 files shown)


bagelface:
Heads-up on two things that touch this branch, neither of which is a request to change it.

**#316 is closed as "keep the docs-side proxy."** That decision is scoped to the REST ingress only and does not pre-empt the CORS policy here. The reasoning there was that `POST /tasks` is behind a bearer API key, so a browser app cannot call it directly without leaking the key, which means CORS on that surface unlocks nothing today. A router used as a wallet RPC URL is the opposite case: it is called from arbitrary dApp origins, so an allowlist is impossible by construction and `allow_origin(Any)` is the correct policy rather than the lazy one. Scoping the layer to `rpc_routes` before the merge, and listing `Authorization` explicitly because the Fetch spec excludes it from a wildcard, both read as right to me.

**#434 removes `tower-http` from main.** Nothing on main references `tower_http`, so it was compiled for nothing. This branch is the only code in any branch that uses it, and it inherited the declaration from main rather than adding it, so a rebase after #434 merges needs to restore two lines:

```toml
# Cargo.toml
tower-http = { version = "0.5", features = ["cors"] }

# router/Cargo.toml
tower-http = { workspace = true }
```

If you would rather not carry that, say so on #434 and I will close it and leave the dependency until this lands. Separately, worth knowing before you pick the rebase back up: main has moved to the v0.3.0 API surface, so `/trigger` is gone and `/tasks` with polling replaced it.

PR #320, opened 2026-07-12
feat(contracts): harden Schnorr settlement — reentrancy guard, in-transition latch, multicall batch, registry packing

RonTuretzky:
Stacks on **#312** (aggregate Schnorr). Closes the three hardening gaps #312 deliberately left open — re-entrancy, midway-state observation, multicall — and pins the gas story down in both execution contexts, with the measurement methodology fixed so the numbers can't be misquoted again.

## Why now

`StateChangeHandlerLib`'s `CALL` update forwards all remaining gas to an arbitrary target **mid-transition** — after `trackState` has bumped the counter, before the transition's later updates land. Two consequences: a `CALL` target holding transition N+1's *valid* quorum signature could re-enter `verifyAndUpdate` (the index check passes — the counter already reads N+1) and interleave two signed transitions; and any external reader called mid-transition observes storage the quorum never signed.

## What's here

**`TransitionGuard` (new)** — EIP-1153 transient-storage guard on both settlement entrypoints (~300 gas, no dirty slot survives the tx, no refund accounting), whose same flag is exposed as `inTransition()` so integrators reading a Gas Killer contract can fail closed on mid-transition state (one warm TLOAD, paid by the reader). `test_latchVisibleMidTransition` demonstrates the exact inconsistency it guards: mid-transition, the counter already reads N+1 while only a prefix of N's writes applied.

**`verifyAndUpdateBatch` (new)** — N independently signed transitions in one tx. Each applied submission is checked exactly as standalone `verifyAndUpdate` (same digest, same registry call) — nothing changes for off-chain signing. **Already-settled indices are skipped** (settlement is permissionless; without the skip, a third party settling one lifted submission standalone would revert the victim's whole batch — found by the adversarial review, PoC'd, fixed, and regression-tested). Index gaps and failing applied sub-transitions revert the batch; the latch is held across it; re-entry through either entrypoint from inside a batch reverts.

ERC-165 compatibility: `ISchnorrGasKillerSDK` is untouched — still single-function, its interfaceId still exactly the `verifyAndUpdate` selector `0x82b35a01` the router preflight probes (`common/src/bindings/mod.rs`). Batch + latch live in a separate `ISchnorrGasKillerSDKBatch`, additionally reported by `supportsInterface`.

**`SchnorrStakeRegistry` packing** — `Operator.weight` (`uint96`, EigenLayer's stake width) packs with `registered`: 3 slots per record instead of 4, saving one cold SLOAD per non-signer where it matters (standalone-tx context). `registerOperator` ABI unchanged; overweight reverts `WeightOverflow`; the boundary value is acceptance-tested (mutation-verified: a `>`→`>=` regression is now caught by the suite).

## Measured gas (forge, solc 0.8.29, prague; registry-attributable)

| `isValidSignature` context | full participation | marginal / non-signer |
|---|---|---|
| **warm** (batch sub-call 2..N, or composed call) | **6,512** | **+4,197** |
| **cold** (standalone tx — what receipts reflect) | **17,033** | **+10,194** (was 12,162 pre-packing) |

Marginal is linear to k=8 (proven on a synthetic 16-operator registry). Warm→cold delta reconciles exactly: cold account (+2.5k) + 4 first-touch base slots (+2k each).

**Measurement finding worth knowing:** `forge test --gas-report` runs calls in *isolation* (fresh EIP-2929 access lists per call) — that's why #312's `6,749` table and the `19,249` a `--gas-report` run prints were **both correct measurements of the same code**, warm vs cold. `SchnorrStakeRegistryGas.t.sol` measures both contexts explicitly, keeps cheatcode calls (asserts/logs are VM calls!) out of the gas brackets, and its warm asserts detect isolation and stand down — the suite is green under plain `forge test` **and** `--gas-report`.

Crossover honesty: cold, 1–2 non-signers already reach the cached-ECDSA registry cost (30,825 @ 3 ops, #311); warm, ~6. Constant-gas is constant *in signers* — the economics assume the healthy-participation regime the coordinator's suspect-exclusion maintains.

## Live-stack integration

The Rust-embedded deploy artifacts (`scripts/bindings/abis/SchnorrStakeRegistry.json`, `SchnorrArraySummationFactory.json`, `common/src/bindings/abis/SchnorrGasKillerSDK.json`) are regenerated — the e2e deploys the registry and example app *from these JSONs*, so they must track the Solidity (byte-parity against a fresh build asserted). `registerOperator`'s signature and the `verifyAndUpdate` selector are unchanged, so the router/deploy Rust needs no code change: `cargo check --workspace` and `clippy -D warnings` green.

## Verification

- **59 forge tests green in both plain and `--gas-report` modes.** 15 new: reentrancy via CALL update through *and from* both entrypoints (attacker try/catches so the guard's revert — `ReentrantTransition` specifically — is the only failure cause), latch visibility, batch skip/front-run/no-op-redelivery/gap-atomicity/per-item signature plumbing (strict mock — the SDK calls the registry via STATICCALL, so expectations are staged rather than recorded)/amortization, uint96 boundary.
- **Adversarially reviewed** by a 5-lens multi-agent pass (guard semantics, batch semantics, packing arithmetic + storage layout, ABI/artifact/e2e parity, measurement integrity), findings verified by independent refutation agents: 8 confirmed, all addressed in this PR; 2 refuted (latch coverage of app-contract functions — pre-existing design; CBOR metadata drift in non-regenerated ECDSA artifacts — pre-existing, cosmetic).

## Deliberately out of scope

- **STOREs-before-CALLs reordering** in `_runStateUpdates` — nearly-free defense-in-depth but a semantic change that must be coordinated with the off-chain EVMSketch encoder (digest parity).
- **ECDSA `GasKillerSDK` port** — same hole (shared `StateChangeHandlerLib`), same two-line `TransitionGuard` adoption, but that file is owned by the #309/#311 branches; patching it here would conflict.
- **Router-side batch submission** (accumulating sequential heights into one tx) — off-chain work; the contract surface is ready.
- **Single signature over a batch digest** (would amortize the verify itself) — changes the signing protocol.

Design notes + full numbers: `contracts/docs/schnorr-hardening-and-gas.md`.

---

## Update — canonical state checkpointing for re-entrant tasks

The guard above *blocks* re-entrancy. This follow-up makes re-entrancy **correct** instead, so a target whose task legitimately makes external/re-entrant calls settles faithfully.

**Problem.** When a task's native execution makes an external call — especially one that re-enters the Gas Killer target — the on-chain settlement must present the target's storage exactly as native execution had it *at that moment*, or the re-entrant read/write observes state that never existed. The legacy encoder emits raw depth-1 `SSTORE`s in trace order and leans on each `CALL` replaying natively; it makes no guarantee about the storage a re-entrant call sees mid-program, and silently corrupts if a re-entrant replay diverges.

**Fix (in companion PR [gas-killer/gas-analyzer#173](https://github.com/gas-killer/gas-analyzer/pull/173)).** A new `compute_state_updates_canonical` tracks the target's storage image across all trace frames (revert-aware journaling, exactly like the EVM) and emits **canonical state slices**: before every external `CALL`/`CREATE` it brings the target's storage to the exact native image at that point (so a re-entrant call reads canonical state), and a final slice re-asserts the canonical end state (so the commit is correct even if a re-entrant replay diverged). It's still a program of the **existing** `STORE`/`CALL`/`LOG`/`CREATE` update types — `StateChangeHandlerLib` applies it with **no contract change**; canonical checkpointing is purely a smarter ordering + slicing. Slices are ascending-sorted and same-slot writes collapse, so the program is byte-deterministic across provers (digest-stable). 11 new encoder unit tests there.

**Service wiring (this PR).**
- `config::state_encoding()` reads `STATE_ENCODING` (`legacy` default | `canonical`), via a pure testable parser that rejects unknown values rather than diverging digests.
- `GasKillerValidator` carries the encoding and calls `call_to_encoded_state_updates_with_evmsketch_mode`. Both the node and router construct via `new()`, which reads `STATE_ENCODING`, so they stay in lockstep — mandatory, since the encoding changes `storage_updates` and therefore the task digest.
- `example.env` documents it; `docker-compose.yml` threads it to all four services (default `legacy`, so the shipped e2e is byte-identical).

**Dependency / merge order.** This PR now pins `gas-analyzer` to the `RonTuretzky/canonical-state-checkpoints` branch. **Merge gas-analyzer#173 first**, then re-pin this PR to the resulting main commit before merging.

---

## Update — re-entrancy safety e2e (`ReentrantCheckpoint`)

A purpose-built re-entrant target proves the canonical path handles re-entrancy safely end-to-end.

- **`ReentrantCheckpoint`** (SchnorrGasKillerSDK example): task `advance()` increments `counter`, makes a **mid-transition external CALL** to `ReentrantObserver`, then writes `lastObserved`. Traced off-chain → canonical program `[Store(counter,N), Call(observe(N)), Store(lastObserved,N)]`, applied by `verifyAndUpdate` with **no contract change**.
- **`ReentrantObserver`**: re-entered mid-transition; its checks are **load-bearing** — it reverts the whole settlement unless the target presents canonical intermediate state (`counter == N`, and the final `lastObserved` write not yet applied). Re-entrant reads hit the unguarded getters, so the `TransitionGuard` doesn't break legitimate re-entrancy while still blocking the cross-transition attack.
- **Foundry** (`ReentrantCheckpoint.t.sol`, 5 tests): the canonical program settles and the observer confirms; **mis-ordered programs are rejected** (call before the counter slice → stale read; final write before the call → looks finalized), proving canonical intermediate state is genuinely required.
- **e2e leg**: new CI matrix `include` — `schnorr` + `STATE_ENCODING=canonical` + `E2E_EXAMPLE=reentrant`. All wiring is gated on `E2E_EXAMPLE` (default `array-summation`), so the existing legs are byte-identical. A green leg proves a re-entrant task settles safely through the live pipeline.


bagelface (changes requested):
Currently, we maintain contracts in a separate repo. This should not be merged into the service repo. Instead it should be migrated to solidity-sdk repo https://github.com/gas-killer/solidity-sdk/pull/58

bagelface:
### Solidity contract changes migrated → [solidity-sdk#61](https://github.com/gas-killer/solidity-sdk/pull/61)

Following the same move that relocated the aggregate-Schnorr contracts to the SDK ([solidity-sdk#58](https://github.com/gas-killer/solidity-sdk/pull/58)), the **contract-side hardening** from this PR now lives in **[gas-killer/solidity-sdk#61](https://github.com/gas-killer/solidity-sdk/pull/61)** (branch `feat/schnorr-settlement-hardening`, into `main`).

**Migrated** (the 4 features named here), with imports remapped to the SDK's `src/schnorr/` layout:

| This PR (`contracts/…`) | solidity-sdk |
|---|---|
| `src/TransitionGuard.sol` | `src/TransitionGuard.sol` (top-level, shared primitive like `StateTracker`) |
| `src/interface/ISchnorrGasKillerSDKBatch.sol` | `src/schnorr/interface/ISchnorrGasKillerSDKBatch.sol` |
| `src/SchnorrGasKillerSDK.sol` (guard + `verifyAndUpdateBatch` + `inTransition()`) | `src/schnorr/SchnorrGasKillerSDK.sol` |
| `src/SchnorrStakeRegistry.sol` (uint96 packing) | `src/schnorr/SchnorrStakeRegistry.sol` |
| `test/SchnorrGasKillerSDKHardening.t.sol` | `test/SchnorrGasKillerSDKHardening.t.sol` |
| `test/SchnorrStakeRegistryGas.t.sol` | `test/SchnorrStakeRegistryGas.t.sol` |
| `test/SchnorrStakeRegistry.t.sol` (uint96 boundary + gas-context note) | merged into the SDK's existing file |

`ISchnorrGasKillerSDK` is untouched — its interfaceId is still exactly the `verifyAndUpdate` selector `0x82b35a01` the router preflight probes. Verified there: **89 forge tests pass**, `forge fmt --check` and `forge build --sizes` clean.

**Not migrated — stays here / in gas-analyzer:**
- The **reentrant-checkpoint example** (`ReentrantCheckpoint`/`ReentrantObserver`/`ReentrantCheckpointFactory` + `ReentrantCheckpoint.t.sol`) and the **`schnorr-hardening-and-gas.md`** design doc. These exist to exercise the off-chain **canonical state encoder** ([gas-analyzer#173](https://github.com/gas-killer/gas-analyzer/pull/173)) end-to-end — a service/e2e concern, not an SDK primitive — so they belong with the e2e wiring here.
- All Rust/service wiring in this PR (bindings, `config.rs`/`validator.rs`, `docker-compose.yml`, `example.env`, `scripts/…`, e2e workflow) is unaffected and stays in this PR.

Once solidity-sdk#61 merges, the embedded deploy artifacts here can be regenerated from the SDK as the source of truth.

_cc @RonTuretzky — flagging in case the merge order (SDK first, then re-point artifacts) matters for this PR._

bagelface:
This PR is stacked on the now-merged Schnorr branch and can't merge as-is (same status as #312 — reference only). Its content has been split and migrated to `main` across the proper repos. Mapping every layer here to where it landed, so this can be closed once the chain below merges:

| Layer in #320 | Migrated to | Status |
|---|---|---|
| Solidity hardening (TransitionGuard, `verifyAndUpdateBatch`, `uint96` registry packing) + Foundry tests | **solidity-sdk#61** (+ review follow-up solidity-sdk#62) | ✅ merged |
| `ReentrantCheckpoint` / `Observer` / `Factory` example + its Foundry test | **solidity-sdk#63** | 🔜 open |
| Regenerated core deploy artifacts (`SchnorrGasKillerSDK` / `SchnorrStakeRegistry` / `SchnorrArraySummationFactory`) | **service#327** (refreshed from solidity-sdk `main`) | 🔜 open |
| Canonical state-encoding wiring (`config`/`validator`/cargo/`docker-compose`/`example.env`) + reentrant e2e (bindings/deploy/`send_request`/`run_e2e`/e2e matrix leg) | **service#332** (stacked on #327) | 🔜 open |
| Canonical off-chain encoder (`compute_state_updates_canonical`) | **gas-analyzer#173** | ✅ merged |
| Router-side batch submission (drive `verifyAndUpdateBatch` end-to-end) | **issue #328** | 📋 tracked, deferred |

Merge order: solidity-sdk#63 → #327 → #332, after which **#320 can be closed** (with #328 left open for the router-batch follow-up).

Note during migration: the SDK's `foundry.toml` now pins `evm_version = cancun`; the service artifacts are kept on `prague` (⊇ cancun for EIP-1153) to match the rest of `main`'s embedded artifacts, so their diff is purely the contract hardening.

bagelface:
Closing this now that https://github.com/gas-killer/service/pull/332 is merged in. Will keep the branch as a reference for now.

PR #322, opened 2026-07-13
refactor: consume commonware-restaking's aggregation-engine library

bagelface:
## Summary

PR #299 vendored a full generic aggregation-engine stack directly into `service` (BN254 `certificate::Scheme`, EigenLayer operator-discovery client, generic EigenLayer contract bindings, task-directive wire format, and the node/router actors driving `commonware-consensus`'s `aggregation::Engine`) because no library provided that integration at the time.

`BreadchainCoop/commonware-restaking`'s `bagelface/integrate-aggregation-engine` branch now provides exactly this as a reusable library (`commonware-avs-{core,node,router,eigenlayer,bindings}`), with a pluggable `Scheme` chassis (BN254 today, plus a new ECDSA scheme). This PR deletes the vendored copies and wires `service` up to the library instead, preserving current behavior exactly — same env vars, same on-chain call shape, same journal/ops layout.

## What's deleted (superseded 1:1 by the library)

- `common/src/{bn254,consensus,eigenlayer,tasks}` and the generic EigenLayer contract bindings
- `node/src/{automaton,task_book,reporter}`
- `router/src/{automaton,reporter,submitter}` and the generic half of `router/src/sequencer.rs`

## What stays (Gas-Killer-specific)

- `GasKillerTaskData`/`build_payload_hash`, `GasKillerValidator`'s EVMSketch execution, the `GasKillerSDK` contract binding, the `verifyAndUpdate` call, HTTP ingress/SQLite API-key store, and all ops layering (healthz, metrics, speculative prebuild, multi-chain wallets)

## Seams added

- `GasKillerValidator` implements the library's `ValidatorTrait`
- `GasKillerHandler` implements `BlsSignatureVerificationHandler`
- New `GasKillerTaskSource` implements `TaskSource` (merges the old `wait_for_task`/`enrich`)

Net: 6,099 lines removed, 330 added.

## Notes

- BN254-only for now; the library's new ECDSA scheme is not wired up (no deployed contract verifies it).
- The git dependency is pinned by branch name (`bagelface/integrate-aggregation-engine`), matching this workspace's existing convention for other `commonware-avs-*`/`sp1-cc-*` git deps. Should be re-pinned to a commit/tag once that branch merges upstream.

## Validation

`cargo fmt --check`, `clippy --all-targets --all-features -D warnings`, `check --all-targets --all-features`, and `test --all-targets --all-features` all pass. The docker-compose e2e (real 3-of-3 BN254 certificate on-chain) has not been re-run in this environment — worth doing before merging.


BreadrichEngels (commented):
Reviewed by checking out the branch and reading the three seams against the code they replace. This is a high-quality refactor — the deletion is clean and the behavior-preservation claim holds where it matters most. Net −6k lines of vendored code for a library dependency is a great trade.

**What I verified:**

- **Digest is byte-for-byte preserved.** `common/src/task_data.rs` has **zero** diff — `build_payload_hash` (the thing the on-chain contract verifies) is untouched. This is the one invariant that absolutely could not drift, and it doesn't.
- **`TaskSource` seam (`sequencer.rs`)** is a faithful lift of the old creator: same explicit-vs-auto `transition_index` path (concurrent `stateTransitionCount` / EVMSketch / `eth_chainId`), same "announce without `storage_updates`" trust model (nodes recompute; digest built from local re-execution), same wire-size validation before broadcast.
- **`BlsSignatureVerificationHandler` seam (`executor.rs`)**: the `verifyAndUpdate` call construction is unchanged — only the binding type-shape adapts (`getNonSignerStakesAndSignatureReturn._0` → the `NonSignerStakesAndSignature` struct directly) and a `round`→`height` rename to match the library's aggregation-engine keying.

**Nice catch buried in the refactor (worth surfacing):**

The validator's digest cache key went from `(transition_index, block_height)` to `(transition_index, block_height, target_address, from_address, value, selector)`. The old key was a **latent collision bug** — two different contracts sharing a `transition_index` at the same `block_height` would collide, and the validator could hand back the *wrong* contract's cached digest. The new key + the `digest_cache_key_distinguishes_different_contracts` test fix that. Good.

One note on this: it's a genuine **behavior change**, not pure refactor, so the "preserves current behavior exactly" framing isn't quite accurate here (it's better than before). Worth a line in the PR description, and it's specifically something the e2e run should exercise.

**Two things to resolve before merge (you flagged both — reinforcing):**

1. **Branch-pinned git dependency.** All five `commonware-avs-*` crates now point at `branch = "bagelface/integrate-aggregation-engine"` — an unmerged branch of another repo. `Cargo.lock` pins a specific rev so *builds* are reproducible, but the branch is mutable (a rebase/force-push invalidates the lock rev) and this makes `main` depend on unreleased upstream work. Land the library branch in `commonware-restaking` first, then re-pin here to a tag or commit SHA. I would not let this reach the mainnet build path branch-pinned.

2. **e2e not re-run.** Swapping the entire aggregation stack for a library, with unit tests green, does not cover the real p2p + 3-of-3 BN254 certificate + on-chain path. The docker-compose e2e is mandatory here before merge — that's exactly where a subtle seam mismatch (or the digest-cache-key change) would show up.

**Minor:**
- The `round`→`height` rename means `dispatch_time` must be *stamped* by the sequencer under the same `height` key the submitter *consumes*. If those ever diverge it's metrics-only (p2p round-trip / round-latency go silently wrong), not a correctness issue — worth a quick sanity check during the e2e.
- `contract_address.map(|a| a.to_string())` vs the old `format!("{:?}", …)` is a cosmetic change to a logged/result field only; on-chain call unaffected.

**Scope caveat:** I verified the three seams, the digest invariant, the executor call shape, and the dep pins. I did **not** deeply verify the "same journal/ops layout" claim in `node/src/main.rs` / `router/src/main.rs` storage config — worth another set of eyes there, since a changed journal partition/path would matter for in-place upgrades.

Solid work — once the dep is re-pinned to a merged commit and the e2e passes, this is a clear improvement.

PR #324, opened 2026-07-17
feat: wire task status transitions and startup re-queue

### LINKED ISSUE #308: Wire task status lifecycle transitions (queued → processing → ready/failed)
## Parent Epic
gas-killer/roadmap#15

## Background

The task model (#300), submission (#193), and status-polling endpoints (#194) are in place, but **nothing transitions a task's status yet.** Accepted tasks are persisted `queued` and stay `queued` forever — the aggregation channel carries `GasKillerTaskRequest` with no task id, so the orchestrator can't correlate the work it processes back to a task row. As a result `GET /tasks/{id}` always reports `queued`, which makes the polling API (#194) not yet useful end-to-end.

## Scope

Thread the task id through the pipeline and drive the lifecycle:

- Change the task channel payload to carry the `task_id` (and `key_id`) alongside the request, so the orchestrator/creator can identify the task it is working.
- On dequeue: `queued → processing` (`update_task_status`).
- On aggregation completion: `→ ready` with the executable payload (**depends on #195** for the payload shape) — until #195, this may settle without a payload or be gated behind it.
- On failure: `→ failed` with the error (`mark_task_failed`).

## Hard sequencing constraint

The startup re-queue (`incomplete_tasks`, a #192 acceptance criterion) **must not** be wired before these transitions exist. Otherwise a restart re-enqueues tasks that already completed and re-runs them. Land transitions first, then the re-queue (they can be the same PR, transitions-first).

## Acceptance Criteria

- [ ] Task id flows through the aggregation channel to the orchestrator
- [ ] `queued → processing → ready|failed` transitions are persisted as a task is worked
- [ ] `GET /tasks/{id}` reflects the live status
- [ ] Startup re-queue wired only alongside/after transitions (satisfies the #192 criterion without re-running completed work)

bagelface:
## Summary

Closes gas-killer/service#308. Progresses gas-killer/service#194 — until now, an accepted task stayed `queued` forever, since nothing threaded the task id into the aggregation pipeline or reported back how it settled.

**Not a straight rebase of #308's original branch.** `bagelface/task-lifecycle-transitions` (the branch #308 was originally scoped against) targeted the pre-migration `Creator`/`ListeningGasKillerCreator`/`BlsEigenlayerExecutor` pattern, which `task-lifecycle-refactor` has since replaced with the `commonware-consensus` aggregation engine (`sequencer.rs` + a rewritten `executor.rs`; `creator.rs`/`builder.rs` are gone). This PR re-implements the same intent — thread a task id through dispatch, mark `processing` on dequeue, settle `ready`/`failed` when the work resolves — against the new sequencer/engine model instead.

## What changed

- **`QueuedTask { task_id, request }`** replaces the bare `GasKillerTaskRequest` on the ingress channel, so `submit_task_handler` hands the task id all the way down to the sequencer.
- **`InFlightTask`** (`router/src/sequencer.rs`) — a single shared slot holding the currently-dispatched task's id, in place of the old round-keyed `HashMap`. The new `Sequencer` (in `commonware-avs-router`) drives exactly one height at a time and only calls `TaskSource::next_task` again once the current task resolves, so a single slot suffices:
  - `next_task` sets the slot when a task starts.
  - `GasKillerHandler::handle_verification` takes it when a certified height executes, settling `ready` (success) or `failed` (with the execution error).
  - If the slot is still occupied the *next* time `next_task` runs, the previous task's height was skipped by the quorum — the one path that resolves a height without ever calling `handle_verification` — so `next_task` settles it as `failed` before starting the next task.
- Enrichment failures (`enrich` erroring, or the enriched task failing `validate()`) now settle the task `failed` with the specific reason instead of just logging and dropping it.
- **Startup re-queue** (`factories::requeue_incomplete_tasks`) — `IngressHandles` now exposes the opened store; on boot, before the sequencer starts, every task left `queued`/`processing` by a previous router life is rebuilt from its persisted request and pushed back through the same channel a fresh `POST /tasks` uses, so it's indistinguishable from new work.

## Scope

Status transitions and restart recovery only. A `ready` task's `payload` stays `null` — the interactive tier submits on-chain itself and carries no user-executable payload yet, which is gas-killer/service#195's job. `expired` is intentionally untouched; that's TTL/sweep work tracked separately in gas-killer/service#196.

## Testing

- `sequencer.rs`: helper persistence (`processing`/`ready`/`failed`), orphaned-slot detection (settles failed, clears slot, no-ops when empty), and an end-to-end `next_task` case (dequeue → mark processing → enrichment fails against an unreachable RPC → mark failed).
- `executor.rs`: `handle_verification` settles the in-flight task failed when execution errors (forced via `task_data: None`, which fails before touching any provider) and clears the slot; no-ops safely when nothing is in flight.
- `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings` (clean), and `cargo test --all-targets --all-features` all pass across the workspace (136 router tests, full workspace green).

### DIFF
--- a/router/src/executor.rs
+++ b/router/src/executor.rs
@@ -1,4 +1,6 @@
 use crate::metrics::MetricsCollector;
+use crate::sequencer::{InFlightTask, in_flight_task, set_task_failed, set_task_ready};
+use crate::store::SqliteStore;
 use crate::task_data::GasKillerTaskData;
 use alloy::network::Ethereum;
 use alloy_primitives::{Address, Bytes, FixedBytes, U256};
@@ -42,6 +44,13 @@ pub struct GasKillerHandler<P> {
     /// applied to every chain. When unset, per-chain defaults apply. Sourced from
     /// `EXECUTOR_RECEIPT_TIMEOUT_SECS`.
     receipt_timeout_override: Option<u64>,
+    /// Durable store used to settle a task's terminal status once its height
+    /// executes. `None` in store-less test/dev harnesses, where the transition is
+    /// skipped.
+    store: Option<SqliteStore>,
+    /// Shared with [`crate::sequencer::GasKillerTaskSource`]; see
+    /// [`crate::sequencer::InFlightTask`].
+    in_flight: InFlightTask,
 }
 
 impl<P: Provider<Ethereum> + Clone + Send + Sync + 'static> GasKillerHandler<P> {
@@ -56,6 +65,8 @@ impl<P: Provider<Ethereum> + Clone + Send + Sync + 'static> GasKillerHandler<P> 
             dispatch_time: Default::default(),
             interface_cache: Arc::new(RwLock::new(HashMap::new())),
             receipt_timeout_override: None,
+            store: None,
+            in_flight: in_flight_task(),
         }
     }
 
@@ -68,6 +79,8 @@ impl<P: Provider<Ethereum> + Clone + Send + Sync + 'static> GasKillerHandler<P> 
             dispatch_time: Default::default(),
             interface_cache: Arc::new(RwLock::new(HashMap::new())),
             receipt_timeout_override: None,
+            store: None,
+            in_flight: in_flight_task(),
         }
     }
 
@@ -92,6 +105,21 @@ impl<P: Provider<Ethereum> + Clone + Send + Sync + 'static> GasKillerHandler<P> 
     /// the per-chain defaults.
     pub fn with_receipt_timeout(mut self, timeout_secs: Option<u64>) -> Self {
         self.receipt_timeout_override = timeout_secs;
+        self
+    }
+
+    /// Attaches the durable store so a settled height advances its task's terminal
+    /// status.
+    pub fn with_store(mut self, store: SqliteStore) -> Self {
+        self.store = Some(store);
+        self
+    }
+
+    /// Shares the in-flight task slot with [`crate::sequencer::GasKillerTaskSource`].
+    /// Must be the same handle the task source is built with, or a settled height
+    /// won't map back to its task.
+    pub fn with_in_flight_task(mut self, in_flight: InFlightTask) -> Self {
+        self.in_flight = in_flight;
         self
     }
 
@@ -412,6 +440,23 @@ impl<P: Provider<Ethereum> + Clone + Send + Sync + 'static> BlsSignatureVerifica
             }
         }
 
+        // Settle the task this height was executing. `GasKillerTaskSource::next_task`
+        // set the in-flight slot when it dispatched this task; taking it here both
+        // records the outcome and clears the slot so a later skipped height is not
+        // mistaken for this one. The task carries no user-executable payload yet
+        // (that arrives with the payload-delivery work), so a successful execution
+        // is recorded as `ready` with the status transition alone.
+        if let Some(store) = &self.store
+            && let Some(task_id) = self.in_flight.lock().ok().and_then(|mut slot| slot.take())
+        {
+            match &result {
+                Ok(_) => set_task_ready(store, &task_id).await,
+                Err(e) => {
+                    set_task_failed(store, &task_id, &format!("verification failed: {e}")).await
+                }
+            }
+        }
+
         result
     }
 }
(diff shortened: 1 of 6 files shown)


BreadrichEngels (approved):
Closes #308 cleanly — task id flows through the channel (`QueuedTask`), `queued→processing→ready|failed` are persisted as the task is worked, `GET /tasks/{id}` will now reflect live status, and — importantly — the startup re-queue lands *in the same PR* as the transitions, transitions-first, so it satisfies the #192 criterion without the "restart re-runs completed work with nothing to settle it" hazard the issue warned about. Nice adaptation to the post-#184 sequencer/engine model rather than a straight rebase.

**I walked the single-slot state machine across every path, and every task settles:**
- **Success** → `next_task` sets the slot *after* enrich/validate pass, so it's set before the `SequencedTask` is returned (no window where `handle_verification` fires before the slot exists); the certified height's `handle_verification` takes the slot and marks `ready`.
- **Enrich/validate failure** → settled `failed` and `continue`, and the slot was never set (step is after validate), so the next loop's `settle_orphaned_task` correctly finds nothing.
- **Execution error** → `handle_verification` takes the slot and marks `failed("verification failed: …")`.
- **Quorum skip** (the one path that resolves a height without ever calling `handle_verification`) → the slot stays occupied, and `settle_orphaned_task` runs at the *top* of the next `next_task` — before `wait_for_task` blocks — so the skipped task settles `failed` promptly when the skip resolves, not deferred until new work arrives.

The mutual exclusion holds too: `handle_verification` clears the slot, so a settled height can't also be orphan-settled, and vice-versa — no double-settle. The `in_flight` Arc is shared correctly (main.rs builds one `in_flight_task()` and hands `.clone()` to both the submitter/handler and the task source; `create_submitter`'s `.with_in_flight_task` overrides the handler's default). Making the status writes best-effort (log, don't propagate) is the right call — aggregation is the real work and a missed transition is recoverable via the re-queue. Good, deterministic tests (forcing failures via `task_data: None` / an unreachable RPC).

**The load-bearing invariant to keep honest:** the whole single-slot model rests on the upstream `Sequencer` driving *exactly one height at a time* and only calling `next_task` again once the current one resolves. That's stated in the doc and it's your sequencer from #184, but it's the assumption that makes "one shared `Option<String>` slot" sound — if that ever pipelines two heights, settlements misattribute. Worth a note in the sequencer itself so a future change there can't silently break task attribution here.

**Known at-least-once edge worth flagging (not blocking — inherent to the persist-processing/re-queue design, and #308-scoped):** a task whose height actually executed on-chain but where the router crashed *before* `set_task_ready` (or the ready-write itself errored) stays `processing`, so the restart re-queues and re-runs it. The on-chain transition-index ordering makes the second apply idempotent (rejected, not double-applied — so no safety issue), but the re-run then errors and the task ends up marked `failed` despite the transition having landed — a false-negative status. The contract idempotency covers the actual harm; the #192 dedup / #196 TTL work (or an on-chain state check before re-running) is where the status accuracy gets tightened. Just naming it so it's a known, accepted edge.

**Minor:** the failure paths are well-covered, but the *success* settlement (`Ok(_) => set_task_ready` through `handle_verification`) isn't unit-tested — it needs a real certificate/chain, so it rides on the e2e. The `set_task_ready` helper is tested directly, and the branch is trivial, so fine — just confirm the docker-compose e2e exercises a task reaching `ready`. `payload` staying `null` (until #195) and `expired` untouched (until #196) match the stated scope.

CI isn't reporting on the head commit yet — confirm `fmt`/`clippy`/the 136 tests are green. Targets `task-lifecycle-refactor` as intended. Approving.

bagelface:
Addressed both points:

- **At-least-once re-queue edge**: documented directly on `requeue_incomplete_tasks` (the false-negative status risk when a crash/store-write-failure lands between a successful `verifyAndUpdate` and the `set_task_ready` write). Not fixed here per your framing — filed as gas-killer/service#325 with two candidate mitigations (on-chain state check before re-running, or extending #197's dedup to cover replay), linked from the doc comment.
- **Success-settlement coverage**: checked — the e2e (`send_request.rs`, what `run_e2e_test.sh`/CI actually runs) only polled on-chain `currentSum`, never `GET /tasks/{id}`, so the `ready` status write wasn't actually asserted anywhere, just incidentally exercised. Added a bounded follow-up poll (`confirm_task_ready`) that asserts the task reaches `ready` after `currentSum` changes, and fails fast (with the recorded error) if it settles `failed`/`expired` instead.

payload-null (#195) and expired-untouched (#196) confirmed as in-scope-as-stated, no change needed there.

PR #327, opened 2026-07-20
test(e2e): regenerate Schnorr deploy artifacts from hardened SDK contracts (solidity-sdk#61)

bagelface:
The reduced, service-side half of [#320](https://github.com/gas-killer/service/pull/320): the Solidity contract hardening (reentrancy guard, in-transition latch, multicall batch, registry packing) has been migrated to the SDK in **[gas-killer/solidity-sdk#61](https://github.com/gas-killer/solidity-sdk/pull/61)**. This PR carries over **only what's necessary to test those contract changes through the live e2e** — which turns out to be just regenerating the embedded deploy artifacts.

## Why this is enough

The e2e deploys the SDK contracts from embedded forge-artifact JSONs (the `alloy::sol!` macro reads ABI + bytecode at compile time). `main` already runs a `schnorr` matrix leg (`signature_scheme: [bls, schnorr]`), and the schnorr deploy path doesn't touch `weight`/`operators()`/`verifyAndUpdateBatch`/`inTransition`. So regenerating the three schnorr artifacts from solidity-sdk#61 makes that existing leg deploy + settle against the **hardened** contracts (guard on the normal settlement path, packed registry) with no other code change.

## Changes (3 files, regenerated)

| Artifact | Delta |
|---|---|
| `common/src/bindings/abis/SchnorrGasKillerSDK.json` | ABI gains `verifyAndUpdateBatch` + `inTransition` (additive; abstract → no bytecode); `interfaceId` unchanged (still the `verifyAndUpdate` selector `0x82b35a01` the router preflight probes) |
| `scripts/bindings/abis/SchnorrStakeRegistry.json` | `uint96` weight packing + `WeightOverflow` (~+118 B; `operators()` getter `weight` now `uint96`) |
| `scripts/bindings/abis/SchnorrArraySummationFactory.json` | child target embeds the guarded SDK (~+719 B) |

Built from solidity-sdk#61 with **optimizer on / runs 200 / solc 0.8.29 / evm prague** to match the existing artifacts, so the diff is purely the contract hardening. `compilationTarget` shifts `src/ → src/schnorr/`, reflecting the SDK as the source of truth.

## Deliberately excluded from #320

- **Canonical-encoding + reentrant-checkpoint machinery** (`config` `STATE_ENCODING`, `validator` wiring, `docker-compose`, reentrant bindings/deploy, the new e2e leg, the `gas-analyzer` branch pin) — that exercises the off-chain canonical encoder ([gas-analyzer#173](https://github.com/gas-killer/gas-analyzer/pull/173)), not this contract migration.
- **Router-side batch submission** (to drive `verifyAndUpdateBatch` end-to-end) is a deferred off-chain follow-up; batch is covered by the SDK's Foundry tests for now. Tracked separately.

## Verification

- `cargo check --workspace` clean; the ABI delta is purely additive, so no binding consumer breaks.
- CI runs the `schnorr` e2e leg against the regenerated artifacts.

## Notes / merge order

- Artifacts are built from the solidity-sdk#61 **branch**; once #61 merges, regenerate from solidity-sdk `main` for byte-parity.
- solidity-sdk's `foundry.toml` currently pins neither optimizer nor solc, so a bare `forge build` there produces *unoptimized* / latest-solc artifacts that would **not** match these. Worth pinning `optimizer`/`solc_version` in the SDK repo as a follow-up so regeneration is deterministic.


bagelface:
Refreshed the three artifacts from **solidity-sdk `main`** (commit `2112cd4`, which includes both gas-killer/solidity-sdk#61 and the review-feedback follow-up gas-killer/solidity-sdk#62) — replacing the versions built from the #61 *branch*, per this PR's own merge-order note.

- Rebuilt with the settings the rest of the service's embedded artifacts use: **solc 0.8.29 / evm prague / optimizer runs 200** (the SDK's `foundry.toml` now pins optimizer + `evm_version = cancun`, but prague is kept here so the only diff vs the previous artifacts is the contract change, and prague ⊇ cancun for EIP-1153).
- gas-killer/solidity-sdk#62 re-touched `SchnorrGasKillerSDK.sol` and `StateChangeHandlerLib.sol` after the #61 branch (solidity-sdk), so this is a substantive refresh (runtime bytecode shifted, not just metadata), not a no-op.
- ABIs unchanged vs the branch-built versions (still additive: `verifyAndUpdateBatch` + `inTransition`; registry `operators().weight` is `uint96`, `WeightOverflow` present); `cargo check -p gas-killer-common -p scripts` clean.

PR #329, opened 2026-07-21
feat: user-executable payload output (replace on-chain execution)

### LINKED ISSUE #195: User-executable payload output (replace on-chain execution)

bagelface:
Closes #195.

Shifts execution to the user for v1 beta: instead of the router custodying a funded key and broadcasting `verifyAndUpdate`, a completed aggregation round is persisted as a structured **bundle** and `GET /tasks/{id}` returns a ready-to-sign **transaction request** the user submits themselves.

## Acceptance criteria
- Round completion → task `ready` with a populated `payload` (previously always `null`).
- Round persisted as a `TaskBundle` (msgHash, quorum/proof, referenceBlockNumber, storageUpdates, transitionIndex, targetFunction, chainId, validity) in a new `tasks.bundle` column (migration `0003`); the payload is rendered from it.
- `payload` is a full `{ to, data, value, chain_id, estimated_gas, valid_until_block }` request. `to`/`value` are server-controlled (`value = 0`, not payable in beta) so a future on-chain fee is a server change, not an integrator client-code change.
- `valid_until_block = reference_block + PAYLOAD_BLOCK_BUFFER` (env, default 50, compile-time-asserted `<= BLOCK_STALE_MEASURE`).
- A stale or consumed bundle (past `valid_until_block`, or on-chain `stateTransitionCount` advanced) returns a typed `409 PAYLOAD_EXPIRED` re-request error rather than calldata that would revert. Backed by a short-TTL freshness cache plus persisting `expired` on first detection so repeat polls skip the chain reads.
- The `Executor` is disconnected from the completion path (bundle written, not broadcast) but **retained and reachable** — `execute_verification` / `execute_schnorr_verification` and the funded wallet provider are the future per-API-key **auto-execute** entry point and share `prepare_*` with the render path so switching between render and broadcast stays a localized branch.

Both BLS and Schnorr completion twins are handled.

## Implementation
- `common/src/payload.rs` — new `PayloadView` / `TaskBundle` / `BundleProof` DTOs.
- `common/src/config.rs` — `PAYLOAD_BLOCK_BUFFER` accessor + default.
- `router/src/executor.rs` — factored into `prepare_bls`/`prepare_schnorr` (shared preflight) → `render_*` (calldata via `.calldata()` + `eth_estimateGas`, no broadcast) and the retained `execute_*` broadcast path; completion handlers render + persist.
- `router/src/store/tasks.rs` (+ migration `0003`) — `bundle` column, `mark_task_ready_with_bundle`, `mark_task_expired`.
- `router/src/{sequencer,factories,error,ingress,main}.rs` — settle helper carries payload+bundle; `PayloadExpired` error; `render_or_reject_payload` freshness check on `GET /tasks/{id}`; startup warn when buffer exceeds the staleness window.
- `scripts/send_request.rs` + `run_e2e_test.sh` — e2e reworked to poll until `ready`, sign+submit the payload with `FUNDED_KEY`, and assert `currentSum` moved.
- `example.env`, Helm values/router-deployment, docker-compose — `PAYLOAD_BLOCK_BUFFER` threaded.

## Testing
- `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings` (zero warnings), `cargo test --all-targets --all-features` — all pass locally.
- New coverage: payload/bundle serde + wire shape, ABI-decode of rendered calldata, the full stale-check matrix (in-window pass, block-past, transition-mismatch, RPC→503, no-providers skip, non-ready, already-expired short-circuit), and store bundle/expired transitions.
- The docker-compose + Anvil e2e (`run_e2e_test.sh`) runs in the `e2e-test.yml` matrix (bls, schnorr).

### DIFF
--- a/.github/workflows/helm-integration-tests.yml
+++ b/.github/workflows/helm-integration-tests.yml
@@ -572,6 +572,9 @@ jobs:
         # The yield distributor is a BLS-interface GasKillerSDK target; the schnorr
         # router only drives SchnorrGasKillerSDK targets, so this l2 subsystem runs
         # in the bls leg only (it is disabled at helm-install time for schnorr).
+        #
+        # This step mirrors what an integrator does: poll the task to `ready`, fetch the
+        # rendered payload, submit it with a funded key, then assert the on-chain effect.
         if: matrix.signature_scheme == 'bls'
         run: |
           echo "=== Triggering Yield Distribution through router ==="
@@ -584,13 +587,12 @@ jobs:
           fi
           L2_RPC="http://localhost:8546"
           ROUTER_URL="http://localhost:8080"
+          # Anvil dev account #0 (funded on the l1 and l2 forks): the account that signs and
+          # submits the router-rendered payload.
+          SUBMIT_KEY="0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"
 
-          # Capture router state before trigger
-          echo "=== Router pod status before trigger ==="
-          kubectl get pods -l app.kubernetes.io/component=router -o wide
           ROUTER_POD=$(kubectl get pods -l app.kubernetes.io/component=router -o jsonpath='{.items[0].metadata.name}' 2>/dev/null)
           echo "Router pod: $ROUTER_POD"
-          echo "Restart count: $(kubectl get pod $ROUTER_POD -o jsonpath='{.status.containerStatuses[0].restartCount}' 2>/dev/null || echo unknown)"
           echo "=== Router logs before trigger (last 50 lines) ==="
           kubectl logs $ROUTER_POD --tail=50 || true
 
@@ -604,11 +606,9 @@ jobs:
           RESOLVER=$(cast call --rpc-url "$L2_RPC" "$YIELD_DISTRIBUTOR" "resolveYieldDistribution()(bool,bytes)" 2>&1)
           echo "resolveYieldDistribution: $RESOLVER"
 
-          # Build distributeYieldGK() calldata
+          # Build distributeYieldGK() calldata, converted to a JSON byte array for the task request.
           CALL_DATA=$(cast calldata "distributeYieldGK()")
           echo "Calldata: $CALL_DATA"
-
-          # Convert calldata hex to JSON byte array
           CALL_DATA_HEX="${CALL_DATA#0x}"
           BYTES_ARRAY="["
           for ((i=0; i<${#CALL_DATA_HEX}; i+=2)); do
@@ -617,7 +617,7 @@ jobs:
           done
           BYTES_ARRAY+="]"
 
-          # Build and send trigger payload
+          # Submit the task to the router
           PAYLOAD="{\"body\":{\"target_address\":\"$YIELD_DISTRIBUTOR\",\"call_data\":$BYTES_ARRAY,\"transition_index\":$INITIAL_COUNT,\"from_address\":\"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266\",\"value\":\"0\",\"block_height\":$BLOCK}}"
           echo "Payload: $PAYLOAD"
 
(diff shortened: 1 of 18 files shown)


BreadrichEngels (approved):
## Review: user-executable payload output (closes #195)

**Verdict: Approve.** Clean, well-tested execution-model shift. I read the freshness gate, the render path, the store/migration, and the config invariants end-to-end; the security-relevant surface (the `GET /tasks/{id}` staleness check) is correct and — importantly — the chain stays authoritative, so a stale payload can at worst produce a `409` re-request or an on-chain revert, never a bad state transition. All 8 checks are green, including both e2e twins (`bls`, `schnorr`) that actually sign+submit the rendered payload with `FUNDED_KEY` and assert `currentSum` moved.

### What I verified

**Freshness gate (`render_or_reject_payload`) matches the on-chain preconditions exactly.** Cheapest-first, short-circuiting:
- **Block window** is checked on **L1** (`ChainRole::L1`), which is right — `valid_until_block` derives from the certificate's L1 reference block, so measuring against an L2 target height would falsely expire every L2 payload. Nicely called out in the comment.
- **Consumed-bundle** check reads `stateTransitionCount()` on the **target** chain (via `detect_contract_chain`) and rejects unless `current_count == bundle.transition_index`. That's precisely the on-chain precondition: `verifyAndUpdate` requires `transitionIndex + 1 == stateTransitionCount()` *after* `trackState` increments, i.e. `stateTransitionCount == transition_index` at submission time. So the check is neither too loose nor too strict.
- **Fail-safe on RPC error:** a provider failure maps to `503 RPC_UNAVAILABLE` and leaves the task `Ready` (verified by `stale_check_maps_rpc_error_to_503`) — a transient RPC blip never permanently expires a valid payload. This is the right call and easy to get wrong.
- **Expiry is persisted** (`mark_task_expired`) so repeat polls short-circuit without a chain read, and the `Expired` short-circuit runs before any provider access. The 6s (~½-block) cache TTL is a sensible poll-deduplication window.
- `409` (not `410`) with a typed `PAYLOAD_EXPIRED` is the right semantic — the task exists, the resolution is to re-request.

**Render path.** `render_bls_payload`/`render_schnorr_payload` fix `value = U256::ZERO` (server-controlled, so a future protocol fee is a server change not an integrator change), build calldata via `.calldata()`, set `valid_until_block = reference_block_number + payload_block_buffer` with `reference = head − 1` (guaranteeing `referenceBlockNumber < block.number` at submit time), and record an *advisory* `estimated_gas`. The retained `execute_*` broadcast path shares `prepare_*`, so the future per-key auto-execute tier is a localized render↔broadcast branch rather than a fork — good structural call.

**Store/migration/DTOs.** `0003` is a plain additive nullable `bundle TEXT` (safe). Bundle-as-durable-unit with the payload rendered from it is the right layering — the read path re-checks freshness without re-deriving. DTO serde/wire-shape, ABI-decode of rendered calldata, the full stale matrix, and store transitions are all covered by focused tests.

### Non-blocking notes

1. **Buffer↔staleness invariant is only *warned* at runtime.** The `const _: () = assert!(DEFAULT_PAYLOAD_BLOCK_BUFFER <= DEFAULT_BLOCK_STALE_MEASURE)` guards the **defaults**, but `payload_block_buffer()` and `block_stale_measure()` read env independently, and `main.rs` only `warn!`s when a deployment sets `PAYLOAD_BLOCK_BUFFER > BLOCK_STALE_MEASURE`. In that misconfig, `valid_until_block` is set past the on-chain staleness window, so the freshness block-window check considers a payload fresh in the interval `(reference + stale_measure, reference + buffer]` while `verifyAndUpdate` would revert `StaleBlockNumber` — i.e. exactly the "calldata that would revert" this PR exists to prevent. Consider **clamping** `payload_block_buffer` to `block_stale_measure()` at startup (or hard-failing) rather than warning, so the invariant holds under any env, not just defaults.

2. **`detect_contract_chain`'s `getCode` isn't cached.** The transition-count and block-number reads are cached, but chain detection runs unconditionally before the count-cache lookup, so the "a poll costs at most one `eth_blockNumber` and one `stateTransitionCount`" claim understates it by one `getCode` per poll. Caching the resolved `ChainRole` per target (or reusing the executor's ERC-165 detection result) would close that and make the cache comment accurate.

3. **L2-only deployments skip the block-window check** entirely (no `ChainRole::L1` provider), relying solely on the transition-count check + the on-chain staleness revert for the window. Since operator-set state lives on L1 a prod router should have an L1 provider anyway, but worth a one-line note in the deploy docs.

4. **`estimated_gas` fallback constant** — advisory and wallets re-estimate, but if any integrator submits it verbatim as the gas limit, make sure the fallback is generous enough to not OOG a real `verifyAndUpdate`.

None of these block. The freshness/on-chain-precondition alignment and the RPC-fail-safe are the parts that mattered most and they're both right. Nice, well-tested migration off the custodial-key execution path.

PR #330, opened 2026-07-22
feat: enforce global queue depth cap with backpressure

### LINKED ISSUE #191: Global queue depth cap with backpressure

bagelface:
Closes #191.

## Summary

Bounds the ingress task queue so a burst of valid requests can no longer grow memory without limit or leave tasks aggregating hours later against long-stale block references.

Much of the scaffolding already existed (`QUEUE_FULL` error code, `ingress_at_capacity` counter, a queue-depth gauge, a `max_queue_depth` field). This fills the four gaps against the acceptance criteria.

## Acceptance criteria

- **`MAX_QUEUE_DEPTH` env var (default 100).** New `max_queue_depth()` config helper (`DEFAULT_MAX_QUEUE_DEPTH = 100`); `create_ingress` now reads it instead of borrowing `p2p_message_backlog()`. Documented in `example.env` and threaded through Helm (`router.ingress.maxQueueDepth` → `MAX_QUEUE_DEPTH`).
- **503 `QUEUE_FULL` with a `Retry-After` estimate.** `ApiError` carries an optional `retry_after_secs`, emitted as a `Retry-After` header. The queue-full response sends ~60s, sized to roughly one aggregation round (the router drains one task at a time, so a slot frees when the in-flight round settles).
- **Atomic check + enqueue (no TOCTOU).** The old path did `load()` → validate/persist → `fetch_add()`, so two concurrent submissions could both pass the check and push past the cap. Replaced with a `QueueSlot` RAII guard: `reserve()` does a single `fetch_update` compare-and-increment (reserves only while under cap), releases the slot on every early-return path, and is committed only once the task is genuinely enqueued.
- **Prometheus gauge `gas_killer_queue_depth`.** Renamed from `gas_killer_task_queue_depth` to match the spec; in-repo Grafana dashboard query updated.

## Tests

Added coverage for the Retry-After header (unit + HTTP integration), atomic reservation reaching the cap then refusing, slot release on drop vs. retention on commit, and slot release when a request is rejected. `cargo fmt`, `clippy --all-targets --all-features -D warnings`, `cargo check`, and `cargo test` all pass across the workspace.

## Note on the metric rename

Renaming `gas_killer_task_queue_depth` → `gas_killer_queue_depth` is a breaking change for any dashboards or alerts defined outside this repo that reference the old name; the in-repo Grafana panel is updated here. Chose to rename rather than alias because the acceptance criteria specified the exact name.

### DIFF
--- a/common/src/config.rs
+++ b/common/src/config.rs
@@ -189,6 +189,25 @@ pub fn p2p_message_backlog() -> usize {
         .unwrap_or(DEFAULT_P2P_MESSAGE_BACKLOG)
 }
 
+/// Default maximum number of tasks the ingress queue holds before shedding load.
+///
+/// The router processes one task at a time, so a deep queue means later submissions are
+/// aggregated long after their block references have gone stale. Capping the depth bounds
+/// both memory and worst-case queue latency; requests arriving at capacity are rejected with
+/// `503 QUEUE_FULL` rather than accepted and starved. Configurable via `MAX_QUEUE_DEPTH`.
+pub const DEFAULT_MAX_QUEUE_DEPTH: usize = 100;
+
+/// Reads the ingress queue-depth cap from `MAX_QUEUE_DEPTH`, defaulting to
+/// [`DEFAULT_MAX_QUEUE_DEPTH`]. Zero or unparseable values fall back to the default,
+/// since a cap of zero would reject every request.
+pub fn max_queue_depth() -> usize {
+    env::var("MAX_QUEUE_DEPTH")
+        .ok()
+        .and_then(|v| v.trim().parse().ok())
+        .filter(|&v: &usize| v > 0)
+        .unwrap_or(DEFAULT_MAX_QUEUE_DEPTH)
+}
+
 /// Reads the P2P channel rate limit from `P2P_MESSAGES_PER_SECOND` and returns the
 /// per-message quota period (`1 / rate`), defaulting to
 /// [`DEFAULT_P2P_MESSAGES_PER_SECOND`] when unset or invalid.
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -11,9 +11,9 @@ pub use config::{
     ChainRole, DEFAULT_PAYLOAD_BLOCK_BUFFER, KeyConfig, OrchestratorConfig, SignatureScheme,
     SpeculativePrebuildConfig, ack_messages_per_second, agg_activity_timeout, agg_window,
     block_stale_measure, detect_chain_for_address, get_operator_states, load_key_from_file,
-    load_orchestrator_config, p2p_message_backlog, p2p_quota_period, payload_block_buffer,
-    quorum_threshold_fraction, rebroadcast_interval, round_timeout, schnorr_messages_per_second,
-    schnorr_stage_timeout, signature_scheme, storage_directory,
+    load_orchestrator_config, max_queue_depth, p2p_message_backlog, p2p_quota_period,
+    payload_block_buffer, quorum_threshold_fraction, rebroadcast_interval, round_timeout,
+    schnorr_messages_per_second, schnorr_stage_timeout, signature_scheme, storage_directory,
 };
 pub use payload::{BundleProof, PayloadView, TaskBundle};
 pub use providers::{build_read_providers, chain_rpc_urls_from_env};
--- a/example.env
+++ b/example.env
@@ -249,6 +249,12 @@ SIGNER_ENDPOINT=http://signer:50051
 INGRESS=true
 INGRESS_ADDRESS=0.0.0.0:8080
 INGRESS_TIMEOUT_MS=0
+# MAX_QUEUE_DEPTH: maximum number of tasks allowed in the ingress queue before the router sheds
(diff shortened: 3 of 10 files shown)


BreadrichEngels (approved):
## Review: global queue-depth cap with backpressure (closes #191)

**Verdict: Approve.** The core of this PR is the TOCTOU fix, and it's correct. Clean RAII design, all four acceptance criteria met, thorough tests, and all 8 checks green (both e2e twins + both Helm charts).

### The concurrency fix is right ✅
The old `load() → validate → fetch_add()` genuinely raced — two submissions could both observe `depth < max` and both increment past the cap. The replacement collapses check-and-increment into one atomic:
```rust
queue_depth.fetch_update(Relaxed, Relaxed, |n| (n < max).then_some(n + 1))
```
`fetch_update` only returns `Ok` when the closure yields `Some`, so the increment happens **iff** we were under the cap at that instant — no window between the check and the bump. `queue_slot_reserves_up_to_cap_then_refuses` pins it.

The RAII `QueueSlot` is the right tool for "release on every early-return path." I checked the accounting balances over the whole lifecycle:
- **reserve** `+1`;
- **reject / validation-fail / `sender.send` error** → `slot` drops uncommitted → `-1` (net 0);
- **success** → `commit()` consumes `self` so `Drop` is a no-op, the `+1` persists, and the sequencer's existing per-dequeue `fetch_sub` balances it (net 0).

Crucially `commit()` is called only **after** `sender.send(...)` succeeds, so there's no path where the slot is committed but the task never reached the channel — and because `commit(self)` moves the guard, the borrow checker guarantees no post-commit drop-release. The `Relaxed` ordering is fine here: the invariant is enforced by a single atomic RMW on one counter, and there's no dependent memory being published through it. `queue_slot_releases_on_drop_but_not_after_commit` and `test_rejected_request_releases_reserved_slot` cover both branches.

### Rest of the criteria
- **`MAX_QUEUE_DEPTH` (default 100).** `max_queue_depth()` filters `v > 0` so a `0`/garbage env can't wedge the service into rejecting everything; `create_ingress` and `IngressState::new` both moved off the borrowed `p2p_message_backlog()`. Threaded through `example.env`, Helm values, and the deployment env. Good.
- **`Retry-After` on 503.** `ApiError.retry_after_secs` → `Retry-After` header (integer delta-seconds, spec-valid); `queue_full(msg, 60)` sizes it to ~one round. `secs.to_string()` is always header-safe so the fallible `HeaderValue::from_str` never actually drops it. Unit + HTTP integration tests both assert it, and `error_without_retry_after_omits_the_header` guards the negative.
- **Metric `gas_killer_queue_depth`.** Renamed + in-repo Grafana panel updated.

### Non-blocking notes
1. **Metric rename is a breaking change for anything outside this repo.** You've flagged it in the description and updated the in-repo dashboard, which is the right call — just make sure the deploy runbook covers updating any **external Prometheus alert rules / Grafana panels** that still reference `gas_killer_task_queue_depth`, since those fail silently (an alert that stops firing looks identical to "healthy"). A release-note/CHANGELOG line would help operators catch it.
2. **`QUEUE_FULL_RETRY_AFTER_SECS = 60` is decoupled from the configured round timing.** It's a static hint while actual drain time is governed by `round_timeout`/`agg_window`; a deployment that tunes those significantly would advertise a stale backoff. Advisory-only, so fine for now, but deriving the hint from the round-timeout config would keep it honest.
3. **Semantic shift worth noting: the cap/gauge now count reserved-but-not-yet-enqueued slots** (requests mid-validation), not just enqueued tasks. That's arguably *more* correct — it bounds concurrent on-chain-validation RPC load and memory too — but it means `gas_killer_queue_depth` transiently reads a bit higher than strictly-queued task count, and a flood of *invalid* requests briefly occupies capacity while they validate and release. Reasonable trade-off; just call it out in the metric's help text or the runbook so it isn't mistaken for a leak.
4. **Gauge accuracy under concurrency (trivia):** the `Gauge::set()` calls compute their snapshot separately from the atomic RMW, so under heavy concurrency the gauge can momentarily lag the true counter. The counter itself is always exact, so cap enforcement is unaffected — purely an observability nuance.

None block. The atomic reservation + RAII release is exactly the right shape for this, and the test matrix hits all the paths that matter. Nice tidy fix.

PR #331, opened 2026-07-23
feat: add per-API-key rate limiting to task ingress

### LINKED ISSUE #190: Per-API-key rate limiting on /tasks

bagelface:
Closes #190.

## Summary

Adds per-API-key request rate limiting to the ingress `POST /tasks` endpoint so a single misbehaving client can no longer saturate the shared task queue and starve other users. Reuses the `governor` crate already vendored for P2P rate limiting.

## Acceptance criteria

- **Configurable global default (`RATE_LIMIT_RPM`, default 60).** `rate_limit_rpm()` config helper; wired through `example.env`, Helm values (`router.ingress.rateLimitRpm`), and the router deployment env.
- **429 with `Retry-After`.** Over-limit requests get `429 TOO_MANY_REQUESTS` + `RATE_LIMITED` + a `Retry-After` header (whole seconds until the next request is allowed), via `ApiError::rate_limited`.
- **Keyed by `key_id`, not IP.** The limiter is keyed by the authenticated API key's id.
- **Per-key override at creation.** New `rpm_limit` column (migration `0004`); `POST /admin/keys` accepts `{"rpm_limit": N}` (rejects `0`); `verify_api_key` returns `AuthenticatedKey { id, rpm_limit }` so auth and quota selection share one lookup.
- **In-memory counters, reset on restart.** `KeyRateLimiter` holds a `DashMap<key_id, governor limiter>`; nothing is persisted except the per-key ceiling.
- **Rejections increment a counter (OPS-1).** `gas_killer_ingress_requests_rate_limited`.

## Design notes

- **Where the check runs:** immediately after authentication and before the queue-slot reservation, so an abusive key is turned away before it can occupy capacity or trigger on-chain validation RPCs.
- **Per-key quotas:** each key gets its own `governor` direct limiter built on first use with its effective quota (override, or the global default) and cached in the `DashMap`. Overrides are fixed at key creation, so the cached limiter stays correct for the key's lifetime.
- **Algorithm:** the issue says "sliding window"; `governor` implements GCRA (leaky-bucket), which is its native mechanism and what the implementation note pointed at. `per_minute(60)` = a burst of 60, then refill one cell/second.
- **Scope:** keyed by `key_id`, so the store-less local dev path (`allow_unauthenticated`, no key) is not rate-limited — unreachable in production.
- **Memory:** the map grows one small entry per distinct key seen since startup. Key ids are admin-minted and bounded, and all state drops on restart, so no eviction is needed for v1.

## Tests

14 new tests: limiter unit tests (burst-then-block, per-key isolation, override precedence, override pinned at first use), store round-trip of `rpm_limit` through create/verify/list, and HTTP-level 429 + `Retry-After` + metric increment + per-key isolation + override widening + admin `rpm_limit` validation. `cargo fmt`, `clippy --all-targets --all-features -D warnings`, `cargo check`, and `cargo test` all pass across the workspace.


BreadrichEngels (approved):
## Review: per-API-key rate limiting on task ingress (closes #190)

**Verdict: Approve.** Clean, correctly-layered, and well-tested — a natural sibling to #330. The concurrency in the keyed limiter is right, the SQL is parameterized, the 429+`Retry-After` reuses the infra from #330, and all four acceptance criteria are met. All 8 checks green (both e2e twins + both Helm charts), 14 new tests.

### What I verified

**Keyed limiter (`rate_limit.rs`) is concurrency-correct.** One `governor` GCRA limiter per `key_id` in a `DashMap`. The `limiter_for` fast/slow path is the part to get right and it is: the `get` read-guard is only held inside the `if let Some(...) { return … }` arm, so when the key is absent no shard guard is alive at the `entry(...).or_insert_with(...)` call — that sidesteps the classic DashMap "hold a `Ref` across a write on the same shard" deadlock. And `or_insert_with` closes the create race (two concurrent first-hits for a new key converge on one limiter, first insert wins). `retry_after` = `wait.as_millis().div_ceil(1000).max(1)`, so a sub-second wait still advertises ≥1s and a client honoring it never retries before the cell refills.

**Ordering is correct.** The check runs after `authenticate_caller` and *before* `QueueSlot::reserve`, so an abusive key is turned away before it can occupy a queue slot or trigger on-chain validation RPCs — rate limiting is the correct outer gate over #330's cap. Only authenticated callers are limited; the store-less dev path (`auth = None`) is skipped, and `key.rpm_limit.and_then(NonZeroU32::new)` defensively falls back to the default even if a `0` ever slipped through (creation already rejects `0`).

**Store + migration.** `verify_api_key` now returns `AuthenticatedKey { id, rpm_limit }` from a single `UPDATE … RETURNING id, rpm_limit`, so auth and quota selection share one lookup — nice, no extra round-trip on the hot path. `rpm_limit` is `.bind()`-ed everywhere (no interpolation), migration `0004` is a plain additive nullable `INTEGER`, and `create_api_key` delegates to `create_api_key_with_rpm(None)` so existing callers are unaffected. `rpm_override_round_trips_through_create_verify_and_list` + the HTTP-level isolation/override/zero-rejection tests cover the surface well.

### Non-blocking / forward-looking notes

1. **Per-process counters ⇒ the limit is per-replica.** State is in-memory per router process, so if ingress is ever horizontally scaled behind a load balancer the effective limit becomes `N × RATE_LIMIT_RPM`. Fine for the single-router v1 (and the doc note says as much), but if you scale out you'll need a shared backend (e.g. Redis GCRA) or the cap silently multiplies. Worth a one-liner in the design doc so it's a conscious constraint.
2. **No eviction is sound *given the current invariants*, but they're load-bearing.** The map grows one entry per distinct authenticated `key_id` and never evicts; that's bounded only because keys are admin-minted and the map clears on restart. Two things ride on that: revoked keys leave a stale (harmless, never-consulted) entry until restart, and if key issuance ever becomes self-serve/unbounded this turns into unbounded growth. If either changes, add TTL/LRU eviction.
3. **Quota pinned at first use is correct *because* `rpm_limit` is immutable at creation** — there's no update path, so the cached limiter can't go stale. If a future "change a key's rpm" endpoint lands, the `DashMap` entry must be invalidated (or the limiter rebuilt), or the override change won't take effect until restart. `override_is_pinned_on_first_use` documents the current behavior nicely.
4. **GCRA vs the issue's "sliding window" wording.** You've flagged the deviation in the description — just make sure the public API docs describe the *burst-then-refill* behavior (`per_minute(60)` = burst 60, then ~1 cell/sec) so integrators don't expect a strict rolling window and get surprised by the initial burst.

None block. The deadlock-safe keyed-limiter construction and the auth→rate-limit→reserve ordering are the parts that mattered most and both are right. Tidy, consistent with #330. Nice work.

bagelface:
Opened new issue for hardening rate limiter when we scale https://github.com/gas-killer/service/issues/333

Issue #328, opened 2026-07-20
Router-side batch submission: accumulate sequential transitions into one verifyAndUpdateBatch


bagelface:
## Context

The contract-side multicall entrypoint **`verifyAndUpdateBatch`** now exists in the SDK ([gas-killer/solidity-sdk#61](https://github.com/gas-killer/solidity-sdk/pull/61)), and its ABI is embedded in the service (`common/src/bindings/abis/SchnorrGasKillerSDK.json`, regenerated in #327). It applies N independently signed state transitions in one transaction, amortizing the per-tx fixed costs (21k intrinsic, the registry's cold account + base-slot warm-up ~10.5k, the SDK's own config-slot cold accesses) across the batch — sub-transitions after the first verify at warm-access prices (~6.5k vs ~17k cold at full participation).

**What's missing is the off-chain half:** the router still submits one `verifyAndUpdate` per transition. This was explicitly deferred as out of scope in [#320](https://github.com/gas-killer/service/pull/320) ("Router-side batch submission ... off-chain work; the contract surface is ready") and is **not** covered by #327 (which only regenerates the deploy artifacts so the existing e2e exercises the hardened contracts).

## Goal

Teach the router to accumulate consecutive, same-target, ready-to-settle transitions and submit them as a single `verifyAndUpdateBatch` call instead of N separate `verifyAndUpdate` txs.

## Notes from the contract surface

- Each submission is a `SchnorrTaskSubmission` — field-for-field identical to the `verifyAndUpdate` args, **same signed digest**. No change to the off-chain signing protocol; the batch is purely a submission-side optimization. Each item still carries its own `(msgHash, s, Raddr, nonSigners)`.
- Transitions apply in ascending `transitionIndex` order, consecutive from the current count.
- **Already-settled indices are skipped** (permissionless-settlement / front-run / redelivery tolerance); an index **gap** or any failing applied sub-transition **reverts the whole batch**; the in-transition latch is held across the batch.
- Batch does **not** amortize per-non-signer costs (each sub-verify pays its own subtraction). Collapsing the verify itself to one signature over a batch digest would, but that changes the signing protocol and is separately out of scope.

## Scope / acceptance

- [ ] Router batches K sequential same-target transitions into one `verifyAndUpdateBatch` submission (K configurable; sensible default + cap).
- [ ] Correct behavior on races: skip already-settled indices, don't submit gaps, handle a batch-level revert by falling back / re-queuing.
[...]

PR #332, opened 2026-07-23
feat: canonical state encoding + re-entrancy checkpoint e2e

bagelface:
The **final piece of closing [#320](https://github.com/gas-killer/service/pull/320)**: the off-chain half of the canonical-state-encoding work + the re-entrancy e2e that exercises it. #320's contract hardening went to solidity-sdk#61 (merged); its embedded-artifact refresh is #327; this PR carries everything else that wasn't covered by either.

> **Stacked on #327** (base `feat/schnorr-hardening-e2e-artifacts`) so the review diff is only this PR's own changes; it re-targets to `main` automatically when #327 merges. Also depends on **solidity-sdk#63** (the `ReentrantCheckpoint` example) for the two embedded reentrant artifacts.

## Canonical state encoding (node + router)

`STATE_ENCODING=legacy` (default) | `canonical`. `canonical` (from [gas-analyzer#173](https://github.com/gas-killer/gas-analyzer/pull/173), merged) emits canonical state slices before every external call + a final re-assertion slice, so a re-entrant call into a Gas Killer target observes the storage native execution would have had.

- `common/src/config.rs`: `state_encoding()` + a pure, tested `parse_state_encoding()` (rejects unknown values rather than diverging digests).
- `common/src/validator.rs`: `GasKillerValidator` carries the encoding and calls `call_to_encoded_state_updates_with_evmsketch_mode`. Both node and router construct via `new()`, which reads `STATE_ENCODING`, so they stay in lockstep — **mandatory**, since the encoding changes `storage_updates` and therefore the task digest.
- `example.env` documents it; `docker-compose.yml` threads it to all four services (default `legacy`, so the shipped e2e is byte-identical).
- **gas-analyzer** is moved forward via `cargo update` (no `Cargo.toml` branch pin — it's already a default-branch dep, and #173 is on `main`); the `commonware-restaking` pin is unchanged.

## Re-entrancy e2e (`ReentrantCheckpoint`)

A purpose-built re-entrant target proves the canonical path settles re-entrancy safely end-to-end. `ReentrantCheckpoint.advance()` increments `counter`, makes a mid-transition external call to `ReentrantObserver`, then writes `lastObserved`; the observer's checks are load-bearing (they revert the settlement unless canonical intermediate state is presented).

- Embedded artifacts `scripts/bindings/abis/ReentrantCheckpoint{,Factory}.json` regenerated from solidity-sdk#63 (solc 0.8.29 / prague / optimizer 200, matching the other artifacts) + binding stubs + `mod.rs`.
- `deploy_array_summation.rs`: an `E2E_EXAMPLE=reentrant` gate in `deploy_schnorr()` deploys the `{observer, checkpoint}` pair (reusing the same registry/operators/JSON wiring) under the scheme-agnostic `arraySummation` alias.
- `send_request.rs`: a `read_progress_value` abstraction (`counter()` for reentrant, `currentSum()` otherwise) + `advance()` vs `sum()` calldata, preserving `main`'s `/tasks` submission + `GET /tasks/{id}` poll-to-`ready`.
- `run_e2e_test.sh`: capture-then-re-export `STATE_ENCODING`/`E2E_EXAMPLE` (same discipline as `SIGNATURE_SCHEME`).
- `e2e-test.yml`: the `signature_scheme` axis becomes an explicit 3-leg `include` — `(bls, legacy, array-summation)`, `(schnorr, legacy, array-summation)`, `(schnorr, canonical, reentrant)`. All gating defaults to the existing behavior, so the two original legs are byte-identical.

## Out of scope (tracked)

Router-side batch submission (driving `verifyAndUpdateBatch` end-to-end) → **issue #328**. Not part of #320's diff.

## Verification

- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo build -p scripts --bins` all clean; `cargo test -p gas-killer-common` green (incl. the two new `parse_state_encoding` tests). `docker compose config` valid; `bash -n` + YAML parse clean.
- The default (legacy + array-summation) legs are unchanged; the new `schnorr + canonical + reentrant` leg is what proves re-entrancy settles safely through the live pipeline.

## Merge order

solidity-sdk#63 → #327 → this PR (re-verify the reentrant artifacts against solidity-sdk `main` once #63 lands) → then #320 can be closed.


BreadrichEngels (approved):
## Review: canonical state encoding + re-entrancy e2e (closes the last of #320)

**Verdict: Approve.** The digest-affecting part — node/router encoding lockstep — is wired correctly, it's fail-closed and default-Legacy so production stays byte-identical, and the `schnorr + canonical + reentrant` e2e leg is green, which is the actual end-to-end proof that re-entrancy settles safely under the canonical encoder. All 9 checks pass. I focused on the consensus-critical surface (encoding lockstep + digest); the e2e scripts are lower-risk harness code validated by the green reentrant leg.

### Lockstep is correct — I traced the wiring
The encoding changes `storage_updates` → the task digest, so node and router must agree or quorum breaks. Verified they do:
- **Both** `node/src/main.rs` and `router/src/main.rs` construct via `GasKillerValidator::new()`, which reads `config::state_encoding()` from `STATE_ENCODING`.
- The node's `.with_validator_metrics(…)` is a builder that mutates **only** `validator_metrics`, so it preserves the env-derived `state_encoding` rather than resetting it — this was the exact trap I went looking for, and it's avoided.
- The two constructors that hardcode `Legacy` (`with_rpc_url`, `with_chain_rpc_urls`) are test-only; neither binary uses them in production.
- `parse_state_encoding` is **fail-closed**: unknown values `panic!` at startup (loud, before serving) rather than silently defaulting one binary to Legacy while the other reads Canonical. Case-insensitive/trimmed, `None`/empty → Legacy, with tests covering every branch incl. the panic. Default Legacy means the shipped legacy legs are byte-identical to `main`.

The digest cache omitting `state_encoding` from its key is fine: the encoding is per-process-immutable (set once from env at construction), so a given process can never produce a stale-encoding cache hit.

### Merge order / CI
The include-only matrix reasoning is right (a bare `include` on a base axis would *merge* into the matching combination rather than spawn a separate leg), and the two original legs are preserved as `(bls, legacy, array-summation)` / `(schnorr, legacy, array-summation)`. The e2e script's capture-then-re-export of `STATE_ENCODING`/`E2E_EXAMPLE` after `source ../.env` correctly stops the example defaults from clobbering the caller's choice — same discipline as `SIGNATURE_SCHEME`.

### Non-blocking / forward-looking

1. **Helm doesn't thread `STATE_ENCODING` (by design here), so enabling canonical in prod is a future, carefully-gated step.** Production (Helm/k8s) has no plumbing for this var, so it defaults to Legacy — safe and byte-identical, good. But when you wire it for a real canonical rollout, it **must** be set on both the node *and* router deployments atomically. A partial rollout (canonical on one, legacy on the other) produces divergent digests and silently kills quorum — the failure mode is "no settlement," not a crash. Worth calling out in the #328 / production-enablement follow-up as an all-or-nothing flip, same as the commonware lockstep discipline.
2. **The canonical encoder version is now consensus-critical.** The `cargo update` pins `gas-analyzer` to a specific rev in `Cargo.lock`, so every node/router build from the committed lock shares one encoder — correct. Just flag that any *future* `gas-analyzer` bump touching `compute_state_updates_canonical` is a fleet-wide atomic-rollout event (differing encoder versions = differing `storage_updates` = differing digests), and builds must always come from the committed lock, never a re-resolve.
3. **Canonical is only exercised on the schnorr leg.** Fine, since the encoding is orthogonal to the signature scheme and the reentrant example is the point — just noting there's no `(bls, canonical)` coverage, so the canonical path's scheme-independence is assumed rather than tested.

None block. The lockstep design (single env var → both binaries via `new()`, metrics builder preserving it, fail-closed parse, default-Legacy) is exactly right, and the green reentrant leg is the proof that ties #320's contract hardening, the #173 encoder, and this pipeline together. Clean close-out.

bagelface:
Thanks for the thorough trace of the encoding lockstep. Opened **#334** to track the canonical **cutover** — flipping the default `legacy`→`canonical`, threading `STATE_ENCODING` through Helm (node + router, atomically), and adding the missing `(bls, canonical)` + Helm-canonical CI coverage. It carries forward all three of your non-blocking notes, including the all-or-nothing flip discipline and the "encoder rev is consensus-critical" constraint.

Minor: your note 1 referenced #328 for the production-enablement follow-up — #328 is actually the router-batch (`verifyAndUpdateBatch`) issue, so I split the cutover into #334 rather than folding it there.

PR #335, opened 2026-07-24
feat: task deduplication for POST /tasks

### LINKED ISSUE #197: Task deduplication

bagelface:
Closes #197.

## Summary

A client that retries after a timeout submits the same logical request twice. Both tasks pass validation, but only the first to reach the chain succeeds — the second's `transition_index` is already consumed. This makes retries safe and idempotent by keying deduplication on `(key_id, target_address, transition_index)`.

- **`POST /tasks`**: if an in-flight (`queued`/`processing`) or `ready` task exists for the same key, target, and transition index, the existing task id is returned with `200 OK` (and `"deduplicated": true`) instead of creating a new task.
- A `failed` or `expired` task is **not** a duplicate — its work must be re-run — so re-submission creates a fresh task (`202 Accepted`).
- The check and creation are **one atomic SQL statement** (`INSERT … SELECT … WHERE NOT EXISTS`). SQLite serializes writers, so a concurrent retry that loses the race observes the winner's committed row, inserts nothing, and gets that task back.
- The response shape is identical to a fresh submission, plus a `deduplicated` field (present on both paths).

## Design note

Deduplication applies **only to an explicit `transition_index`**. An `"auto"` (NULL) request leaves the slot for the server to resolve at dequeue time, so two auto submissions are distinct requests that each take their own slot (the existing *safe parallel submissions* behavior) and are never collapsed. This also matches the failure scenario in the issue, which is specifically about an explicit index being consumed on-chain.

## Changes

- `router/migrations/0005_task_dedup_index.sql` — partial index on `(key_id, target_address, transition_index)` where the index is non-NULL, serving the dedup lookup.
- `store/tasks.rs` — `SubmittedTask`, `create_task_deduplicated()`, and `active_duplicate_task()`.
- `ingress.rs` — `deduplicated` response field; handler returns `200 OK` without re-enqueue on a dedup hit (releasing the reserved queue slot).
- `metrics.rs` — `gas_killer_ingress_requests_deduplicated` counter.

No new runtime config, so `example.env` / Helm / config fixtures are unchanged.

## Tests

- **Store**: collapses onto queued/processing/ready; re-submits after failed/expired; per-key scoping; distinguishes `transition_index` and `target_address`; auto never dedups; and a 16-way concurrency test on a real file-backed WAL store proving exactly one row is created.
- **HTTP**: `200` + no re-enqueue on retry, `202` + fresh task after failure, and the metric increment.

Local gate green: `cargo fmt --check`, `clippy --all-targets --all-features -D warnings`, `cargo check`, and the full test suite.

### DIFF
--- a/router/migrations/0005_task_dedup_index.sql
+++ b/router/migrations/0005_task_dedup_index.sql
@@ -1,0 +1,16 @@
+-- Serves the deduplication lookup on POST /tasks.
+--
+-- A client that retries after a timeout submits the same logical request twice. An in-flight
+-- (`queued`/`processing`) or `ready` task sharing a submission's
+-- (key_id, target_address, transition_index) is that same work, so the submit path collapses onto
+-- it rather than creating a duplicate that would race a doomed second transaction onto the chain
+-- (only one can consume a given transition_index). A `failed` or `expired` task is not a duplicate,
+-- so re-submission after one proceeds normally.
+--
+-- The index is partial on `transition_index IS NOT NULL`: an "auto" request (NULL index) leaves the
+-- slot for the server to resolve at dequeue time, so two auto submissions are distinct requests
+-- that each take their own slot (safe parallel submissions) and are never deduplicated. The lookup
+-- therefore only ever probes non-NULL index rows, which this index covers.
+CREATE INDEX idx_tasks_dedup
+    ON tasks (key_id, target_address, transition_index)
+    WHERE transition_index IS NOT NULL;
--- a/router/src/ingress.rs
+++ b/router/src/ingress.rs
@@ -3,7 +3,7 @@ use crate::metrics::MetricsCollector;
 use crate::rate_limit::KeyRateLimiter;
 use crate::sequencer::{QueuedTask, TaskQueueDepth, TaskSender};
 use crate::store::{
-    ApiKeyMetadata, AuthenticatedKey, CreatedApiKey, SqliteStore, Task, TaskStatus,
+    ApiKeyMetadata, AuthenticatedKey, CreatedApiKey, SqliteStore, SubmittedTask, Task, TaskStatus,
 };
 use alloy_primitives::{Address, U256};
 use alloy_provider::Provider;
@@ -618,11 +618,17 @@ impl GasKillerTaskRequest {
 }
 
 /// Body returned by `POST /tasks` (and its deprecated alias `POST /trigger`) when a task is
-/// accepted: the id the client polls for status, and the task's initial state.
+/// accepted: the id the client polls for status, and the task's current state.
+///
+/// `deduplicated` is `true` when the submission collapsed onto an existing in-flight or `ready`
+/// task rather than creating a new one — a retry keyed on `(key_id, target_address,
+/// transition_index)` is idempotent, returning the original task id with a `200 OK`. A fresh
+/// submission carries `false` and a `202 Accepted`.
 #[derive(Debug, Serialize, Deserialize)]
 pub struct TaskAcceptedResponse {
     pub task_id: String,
     pub status: TaskStatus,
+    pub deduplicated: bool,
 }
 
 /// `Retry-After` estimate (seconds) returned alongside a `503 QUEUE_FULL`. The router drains
@@ -789,13 +795,38 @@ pub async fn submit_task_handler(
     let store = require_store(&state)?;
     let key_id = auth.ok_or_else(ApiError::unauthorized)?.id;
 
-    let task = store
-        .create_task(&key_id, &request.body)
+    let SubmittedTask { task, deduplicated } = store
+        .create_task_deduplicated(&key_id, &request.body)
         .await
         .map_err(|e| {
             tracing::error!(error = %e, "failed to persist task");
             ApiError::internal("Internal error: failed to persist task")
         })?;
+
+    // A retry that collapsed onto an existing task must not be enqueued a second time — the
+    // original submission already owns a queue slot (or has settled). Release this request's
+    // reserved slot (the guard drops uncommitted) and hand back the existing task with `200 OK`,
+    // so the client polls the same id idempotently.
+    if deduplicated {
+        info!(
+            task_id = %task.id,
+            status = ?task.status,
+            target_address = %request.body.target_address,
+            transition_index = ?request.body.transition_index,
+            "Task deduplicated onto existing submission"
+        );
+        if let Some(m) = &state.metrics {
+            m.ingress_deduplicated.inc();
+        }
+        return Ok((
+            StatusCode::OK,
+            Json(TaskAcceptedResponse {
+                task_id: task.id,
+                status: task.status,
+                deduplicated: true,
+            }),
+        ));
+    }
 
     info!(
         task_id = %task.id,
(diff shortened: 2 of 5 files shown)


BreadrichEngels (approved):
## Review — task deduplication for `POST /tasks`

**Approve.** The idempotency key and the concurrency mechanism are both correct, the test coverage is genuinely thorough (including the one test that matters most — a real file-backed WAL race), and this sits on the ingress/queue layer, not the digest or verification path, so there's no consensus surface. One robustness edge is worth addressing but doesn't block.

### The mechanism is correct

- **The atomic `INSERT … SELECT … WHERE NOT EXISTS` is the right primitive**, and I walked the race: SQLite serializes writers, so the loser's `NOT EXISTS` subquery runs under the write lock *after* the winner commits, sees the winner's row, inserts nothing, gets `None`, and reads the winner back as the deduplicated result. No window between the check and the insert because they're one statement. The 16-way test deliberately uses `connect_at` (file-backed, multi-connection WAL) rather than the in-memory store — which caps at one connection and would serialize the racers, hiding exactly the concurrency this needs to survive. That's the correct call and the reason I trust the green result.
- **The `auto`/NULL exemption is right** and consistently enforced: the explicit-index guard clause returns early to `create_task`, the partial index is `WHERE transition_index IS NOT NULL`, and `auto_transition_index_never_deduplicates` pins that two autos persist two rows. Matches the "safe parallel submissions" behavior and the issue's actual failure mode (an explicit index consumed on-chain).
- **The terminal-status exclusion is correct** — `failed`/`expired` aren't duplicates, so their work re-runs; `queued`/`processing`/`ready` collapse. Covered from both directions.
- **REST semantics are clean**: 200 + `deduplicated:true` for the idempotent retry vs 202 + `false` for fresh, the response field is additive (backward-compatible), and `ingress_accepted == 1` / `ingress_deduplicated == 1` confirms the retry neither re-enqueues nor double-counts. The partial index correctly serves the lookup (the `(key,target,index)` tuple narrows to ~one row before the status filter, so not indexing status is fine).

### One robustness edge worth fixing (non-blocking)

The `None` branch reads the blocker back via `active_duplicate_task`, and if that returns `None` it surfaces a 500 (`"deduplication guard blocked the insert but no active duplicate was found"`). That's reachable: between the loser's blocked INSERT and its read-back, the blocking task can transition `processing → failed/expired`, at which point `active_duplicate_task` (which filters to active statuses) finds nothing. And it's *correlated*, not random — the client is retrying precisely because aggregation is slow, so the in-flight task failing around retry time is exactly the likely case. The blast radius is small (a transient 500; no duplicate is enqueued, so no safety issue) and it self-heals — the client's next retry sees the blocker as terminal and gets a fresh task. But the cleaner behavior is already implied by your own state machine: if the guard blocked but no active duplicate remains, the blocker went terminal, so the right answer is a *fresh* task, not an error. Consider retrying the whole `create_task_deduplicated` once on that specific `None`-after-block case (the re-run's `NOT EXISTS` now passes) instead of erroring. It turns a spurious 500 in the aggregation-failing path into the correct outcome.

### Minor

- **The explicit-index INSERT reimplements `create_task`'s column list** rather than sharing it. A future column added to `create_task` (nullable-with-default especially) would silently diverge here without a test failing, since only a missing NOT NULL column would surface. Worth factoring the shared insert, or at least a comment tying the two together.
- **The single-active-per-`(key,target,index)` invariant holds only if every explicit-index insert goes through `create_task_deduplicated`.** A stray direct `create_task` with an explicit index would bypass the guard. Not the case today (the handler routes through the dedup path), but a one-line note on `create_task` discouraging direct explicit-index use would keep it that way.

Clean, correct, and the concurrency test is exactly the one I'd have asked for. Approving — the read-back 500 is the only thing I'd genuinely like tightened, and it's a follow-up, not a blocker.

BreadrichEngels (approved):
## Re-review at `aab2fbd` — all three notes addressed, approval stands

Verified the fix commit against source, not just the message. Every point from my prior review is handled, and the one that mattered — the read-back 500 — is fixed correctly and safely bounded. CI is fully green (all e2e legs incl. the unbounded one, both Helm legs, clippy).

### The read-back race is now handled the right way

`create_task_deduplicated` wraps the guarded insert in a bounded `for _ in 0..MAX_INSERT_ATTEMPTS` (3) loop:
- **Insert succeeds** (`Some`) → fresh task, as before.
- **Blocked** (`None`) → read back the active duplicate; if found → `deduplicated: true`, as before.
- **Blocked but no active duplicate** (the terminal race I flagged) → loop and re-attempt. On the re-run the blocker is now terminal, so `NOT EXISTS` passes and the insert creates the fresh task the state machine implies — instead of a spurious 500.

I checked the properties that make this safe:
- **Bounded** — three attempts then `anyhow::bail!` with a descriptive message, so a pathologically churning slot can't spin forever. That's the right call; the non-adversarial terminal race resolves in two attempts (block → terminal → re-insert), so 3 is a clean margin, and exhausting it requires sustained concurrent insert→terminal churn on the same slot, where surfacing an error the client retries is acceptable.
- **Fresh `Uuid::new_v4()` per attempt**, so no id is reused across iterations.
- The behavior is exactly what the state machine implies: a blocked insert with no surviving active duplicate means the work must re-run, so a fresh task is the correct answer — not an error.

### The two minor notes are handled too

- **`TASK_INSERT_COLUMNS`** is now a shared const used by both `create_task` and `create_task_deduplicated`, so a future column lands in both inserts at once rather than diverging silently. `create_task` correctly binds `transition_index.map(|i| i as i64)` (NULL for auto) while the dedup path binds the unwrapped `Some`, so both stay consistent.
- **`create_task`'s doc** now states explicit-`transition_index` submissions must route through `create_task_deduplicated` so the single-active-per-`(key,target,index)` invariant keeps holding — closing the "stray direct insert bypasses the guard" gap at the documentation level.

### One small residual (non-blocking)

The new terminal-race retry branch itself isn't directly unit-tested — the existing dedup suite covers the happy paths and the 16-way WAL concurrency race, but not specifically "blocker goes terminal between the insert and the read-back → loop → fresh task." That branch is genuinely awkward to hit deterministically (it needs a seam to force a status transition mid-method), so I don't consider its absence blocking — the logic is simple and the bound is safe. If you want it covered without a timing hack, a thin injectable hook (or a `#[cfg(test)]` callback between the insert and the read-back) would let you assert the loop produces a fresh task rather than erroring. Purely optional.

Clean fix — the bounded retry is exactly the shape I was hoping for, and sharing the column list plus the routing note close the maintainability gaps. Approval stands.

PR #336, opened 2026-07-28
deploy: capture live LLM demo drift and fix scheduling pins

bagelface:
Captures the state of the live LLM demo cluster in git. Everything here was applied to `gas-killer-testnet` by hand on 2026-07-27 during an outage recovery and cost pass, and existed nowhere in the repo — so a `helm upgrade` would have reverted it.

All changes are verified against the live cluster: `kubectl diff` is clean for both manifests, and a full 0.6B ask settled on Sepolia afterwards.

## Scheduling — the outage fix

`default-live-overrides.yaml` pinned the router and nodes to a VM **hostname**, and the committed hostname was already stale, so the file as committed could not schedule. Both pins now select the node **pool** (`role: ops-highmem`), which survives VM recreation and pool upgrades.

Root cause of the 2026-07-27 outage: with the selector stripped, `node-1` landed on the sim pool while the RWO `gas-killer-shared-data` PVC was attached to `ops-highmem`. It deadlocked in `Init:0/2` with no events, which crashlooped the router's `wait-for-nodes` init container 75 times.

Consequently the `demo-expiry` teardown no longer unpins the operators — that step is what caused the deadlock. It also targeted `sim-pool-c4`, which no longer exists, so a teardown would have left the renamed `sim-pool` running at ~$144/mo. Its script is now a readable block scalar instead of an escaped one-liner.

## Overlay convergence

`qwen-fork-refresh` fired an unconditional `anvil_reset` every 8 minutes, wiping every overlay chunk written so far. The 0.6B overlay is 24,364 chunks and was loading at ~4 setCode/s (~100 min), so **it could never converge** — and had not, for as long as that cron had been enabled.

The ensurer now uses a pooled `requests.Session` and `B=100` batches. Measured on the live cluster:

```
OVERLAY LOADED: 24364 chunks in 3s     (~8,980 chunks/s, was ~4/s)
```

The bottleneck was entirely per-request overhead, not anvil. The cron stays suspended anyway: it ignores in-flight traces and duplicates the ensurer's drift-aware refork (`drift > 150 blocks AND inflight() == 0`).

## Cost

anvil requested 5 vCPU but measures ~0.03 vCPU / ~2Gi. That request was the only reason the sim pool needed an 8-vCPU machine; it is now `1500m/3500m` and the pool is a `c4-standard-4` ($144/mo, was $248).

## Drift `kubectl diff` caught

Two cases where applying the committed file would have broken the demo:

- The `gk-shard-bridge` ingress was missing all five live annotations, including `enable-cors` and `cors-allow-origin`. The frontend is on GitHub Pages, so every `/bridge` call is cross-origin. `ssl-redirect: false` also has to stay or nginx 308-loops.
- A committed `restartedAt` annotation was *older* than the live one, so applying would have rewound it and triggered a needless bridge rollout.

## Verification

`verify-artifacts.yaml` adds a repeatable keccak check for the PVC weights. Nothing verified them before: the original staging Job explicitly skipped the hash ("keccak256 via pysha3 not present… trust size"), and nodes mmap the overlay lazily at first analysis, so a healthy fleet proves nothing. Both manifests were recomputed byte-for-byte and matched — 35B `0x7bdf4876…09f01fa9`, 0.6B `0x23216cb9…c4a7ae9`.

The runbook is rewritten: no hostname re-pin step, correct pool names, and the full re-staging + verification procedure.

## Testing

Three end-to-end asks, all returning a byte-identical `pipeline_root` `0xe718…4e0c` across a sim-pool machine swap, a default-pool consolidation, and the overlay fix:

| Run | Context | infer_seconds |
| --- | --- | --- |
| 1 | pre-optimization, cold JIT | 614.5 |
| 2 | post pool changes | 173.1 |
| 3 | post overlay convergence | **113.7** |

On-chain `stateTransitionCount` advanced 22 → 27 across the session. Endpoints `/healthz`, `/forkhead`, `/bridge/healthz` all 200.

Node pools are declared in gas-killer/infra `terraform/llm-testnet` (separate PR); the `role` labels here are the contract with that config.

### DIFF
--- a/bridge/k8s.yaml
+++ b/bridge/k8s.yaml
@@ -19,8 +19,6 @@ spec:
     type: RollingUpdate
   template:
     metadata:
-      annotations:
-        kubectl.kubernetes.io/restartedAt: '2026-07-16T15:15:38-04:00'
       labels:
         app: gk-shard-bridge
     spec:
@@ -103,6 +101,16 @@ kind: Ingress
 metadata:
   name: gk-shard-bridge
   namespace: default
+  annotations:
+    # The frontend is served from GitHub Pages (llm.gaskiller.xyz), so every
+    # /bridge call is cross-origin — without these the browser blocks it.
+    nginx.ingress.kubernetes.io/enable-cors: "true"
+    nginx.ingress.kubernetes.io/cors-allow-origin: "*"
+    nginx.ingress.kubernetes.io/cors-allow-methods: "POST, GET, OPTIONS"
+    nginx.ingress.kubernetes.io/cors-allow-headers: "Authorization,Content-Type"
+    # Must stay "false": TLS terminates upstream, and enabling the redirect makes
+    # nginx 308-loop (same failure the gas-killer-extras ingress documents).
+    nginx.ingress.kubernetes.io/ssl-redirect: "false"
 spec:
   ingressClassName: nginx
   rules:
@@ -177,13 +185,16 @@ spec:
   schedule: '*/8 * * * *'
   successfulJobsHistoryLimit: 2
   # SUSPENDED (2026-07-27) and superseded by qwen-overlay-ensurer's own refork.
+  #
   # This job fires an UNCONDITIONAL anvil_reset, which re-forks from upstream and
-  # therefore wipes every overlay chunk setCode'd into the fork so far. Loading
-  # the 0.6B overlay is 24,364 chunks at ~4 setCode/s (~100 min), so an 8-minute
-  # reset meant the overlay could never converge. It also ignores in-flight
-  # traces, so it can reset the fork mid-trace. The ensurer already reforks
-  # correctly: only when drift > 150 blocks AND inflight() == 0, and it reloads
-  # the overlay afterwards. Re-enable only if the ensurer is removed.
+  # so wipes every overlay chunk setCode'd into the fork. That used to be fatal:
+  # the 0.6B overlay loaded at ~4 setCode/s (~100 min for 24,364 chunks), so an
+  # 8-minute reset meant it could never converge. Batching + connection reuse in
+  # the ensurer cut that to ~3s, so a reset is no longer unrecoverable — but this
+  # job is still the wrong tool. It ignores in-flight traces and can reset the
+  # fork mid-trace, and it duplicates logic the ensurer already does properly:
+  # refork only when drift > 150 blocks AND inflight() == 0, then reload.
+  # Re-enable only if the ensurer is removed.
   suspend: true
 ---
 apiVersion: batch/v1
--- a/bridge/k8s.yaml
+++ b/bridge/k8s.yaml
@@ -176,7 +176,15 @@ spec:
           terminationGracePeriodSeconds: 30
   schedule: '*/8 * * * *'
   successfulJobsHistoryLimit: 2
-  suspend: false
+  # SUSPENDED (2026-07-27) and superseded by qwen-overlay-ensurer's own refork.
+  # This job fires an UNCONDITIONAL anvil_reset, which re-forks from upstream and
+  # therefore wipes every overlay chunk setCode'd into the fork so far. Loading
+  # the 0.6B overlay is 24,364 chunks at ~4 setCode/s (~100 min), so an 8-minute
+  # reset meant the overlay could never converge. It also ignores in-flight
+  # traces, so it can reset the fork mid-trace. The ensurer already reforks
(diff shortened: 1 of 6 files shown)


BreadrichEngels (commented):
## Review: capture live LLM demo drift + scheduling fixes

Different beast from the recent contract PRs — this is deploy/infra drift-capture (manifests, runbook, a keccak-verify Job) plus a fmt-only Rust reflow. It's verified against the live cluster (clean `kubectl diff`, e2e asks settled) and **Ron is the requested reviewer on his own branch (`ron/sharded-inference`)**, so I'm treating this as a supporting read — manifest internal-consistency and the code delta — and deferring the operational sign-off to Ron, since the real validation is against a cluster I can't touch. No blocking findings; a couple of doc-accuracy nits.

### Reads correctly / internally consistent ✅
- **Hostname → pool pin** (`default-live-overrides.yaml`, both router and node): `kubernetes.io/hostname: gke-…-nmbc` → `role: ops-highmem`. This is the right durability fix — a pool label survives VM recreation where a hostname pin goes stale, and the RWO `gas-killer-shared-data` co-location requirement (router + all 3 nodes on one VM) is correctly preserved.
- **`demo-expiry` no longer unpins operators** — the root-cause fix. Removing the `remove /nodeSelector` step is exactly what stops `node-1` landing on the sim pool while the RWO PVC is on ops-highmem, and the inline comment records why. The `sim-pool-c4` → `sim-pool` rename and the escaped-one-liner → block-scalar conversion are consistent with the rest, and the delete/shrink/self-suspend flow (break on `"name"` or `notFound`, 60× retry) is sound for a teardown cron.
- **`qwen-fork-refresh` suspended** with a clear rationale (unconditional `anvil_reset` wiping overlay chunks), superseded by the ensurer's drift-aware refork. Correct call.
- **Bridge ingress CORS + `ssl-redirect: false`** — correct for a GitHub-Pages frontend hitting `/bridge` cross-origin, and the `ssl-redirect` 308-loop note matches the documented `gas-killer-extras` behavior. Removing the stale (older) `restartedAt` annotation avoids a needless rollout — good catch by the diff.
- **anvil rightsizing** (`5`→`1500m` req / `7500m`→`3500m` lim) and the **overlay ensurer** pooled `requests.Session()` + `B=100` batching are the real convergence fix; the ~49KB-hex-per-chunk / ~5MB-body math checks out, and progress logging addresses the "silent multi-minute load" that hid the non-convergence.
- **`verify-artifacts.yaml`** is nicely done: RWO-aware (runs on `ops-highmem`, mounts read-only), and the **`keccak256("") == c5d2…a470` self-test guarding against `hashlib.sha3_256`** is exactly the right defense — keccak-vs-NIST-SHA3 padding is a classic silent-wrong-digest trap. `manifest = keccak(keccak(weights) ‖ keccak(tokenizer))` + the manifest.json cross-check are correct.
- **`common/src/shard.rs`** is a pure rustfmt/clippy reflow — I checked the one non-cosmetic line, `write!(w, "{}\n", …)` → `writeln!(w, "{}", …)`, which is byte-identical output (the clippy `write_with_newline` lint). No behavior change.

### Nits (non-blocking)
1. **Doc drift on the refork threshold.** The ensurer's `REFRESH_AT` is now **60**, but two prose spots still say **150**: the README `deploy/qwen-overlay-ensurer` row ("drift exceeds 150 blocks") and the `qwen-fork-refresh` suspend comment in `bridge/k8s.yaml` ("refork only when drift > 150 blocks"). Since the whole point of this PR is capturing accurate state — and the `REFRESH_AT` comment itself warns that >~100 silently breaks 35B — these should read 60 so the next operator doesn't "fix" it back up.
2. **`cors-allow-origin: "*"`.** Fine for a public demo whose bridge holds a server-side `GK_KEY` (and `*` disables credentialed/cookie CORS anyway), but it does let any origin call `/bridge`. If the bridge ever fronts anything non-public, scope this to the GitHub-Pages origin rather than `*`. Just flagging; it matches the live state you're capturing.
3. Minor: the fmt-only `shard.rs` change is unrelated to the drift capture — harmless, just noting it rode along.

### CI
`Format, Clippy, Check` is green; the two Docker builds are still in progress. Given the only Rust delta is formatting they should pass, but worth confirming green before merge.

Solid capture — the operators-not-unpinned fix and the keccak self-test are the two things I'd have most wanted here. Leaving the authoritative infra sign-off to Ron.

bagelface:
Closing: the LLM demo this captured drift for has been retired, and its cluster pools were removed (gas-killer/infra#9). `ron/sharded-inference` is no longer deployed anywhere.

PR #337, opened 2026-07-28
feat: thread the Schnorr registry notice window through deployment

bagelface:
Service-side half of the `SchnorrStakeRegistry` notice window. Pairs with **gas-killer/solidity-sdk#69** — that PR changes the registry ABI, so these two must land together or the next artifact refresh breaks the Schnorr deploy.

**Merge order: this PR first (or simultaneously), then solidity-sdk#69.** This branch works against both the old and new contract only in the sense that it compiles; the regenerated bytecode it deploys is the new registry, so it needs #69 merged for the source of truth to match.

## What changed upstream

solidity-sdk#69 adds a notice window to `SchnorrStakeRegistry`: a change is announced first and applies only after `noticeWindow` blocks, so `nextPossibleMutationBlock()` gives an off-chain round a horizon it can check rather than assuming the operator set is quiescent. Two ABI-breaking consequences:

1. The constructor takes a 4th argument, `_noticeWindow`.
2. The `operators` getter returns 5 fields — `exitBlock` is appended, because a committed exit now leaves a tombstone instead of deleting the record.

## Changes here

**`scripts/bindings/abis/SchnorrStakeRegistry.json`** regenerated from solidity-sdk `feat/schnorr-operator-set-notice-window`. `sol!` builds the bindings from this artifact at compile time, so no hand-edit of `schnorrstakeregistry.rs` is needed.

The ABI diff is purely additive apart from the two shape changes above — nothing removed. It also picks up `deregisterOperator` / `OperatorDeregistered`, which this artifact predated (solidity-sdk#64).

**`common/src/config.rs`** gains `schnorr_notice_window()` reading `SCHNORR_NOTICE_WINDOW`, with `DEFAULT_SCHNORR_NOTICE_WINDOW = 0`, following the `quorum_threshold_fraction` pattern (parse helper + unit tests, tolerant of missing/empty/unparseable values).

**`scripts/deploy_array_summation.rs`** passes it as the 4th constructor argument and logs it.

**`example.env`** documents the variable alongside the other immutable registry constructor params.

## Why the default is 0

Zero applies changes as soon as they are committed, i.e. today's behaviour. That is correct for this stack rather than merely convenient: `deploy_schnorr` registers the entire operator set *before* the target contract deploys — the existing comment on that loop already notes the ordering matters because `effectiveBlock` advances on every registration — so no round is ever in flight for a mutation to invalidate, and a non-zero window would only make the e2e wait.

The env var is what makes a real deployment able to choose otherwise. `example.env` records the sizing rule: above the signing round's duration plus `BLOCK_STALE_MEASURE`, since a rendered payload stays submittable for that long. The registry cannot enforce it — the staleness bound lives on each consumer contract and is adjustable there, while one registry may back several consumers.

## What I did not change

**No Helm values.** The Schnorr registry is deployed by the local `deploy_array_summation` binary in both `run_e2e_test.sh` and `run_helm_test.sh`; no chart template constructs it (`helm/` contains no `SCHNORR` reference at all). A Helm value would be dead config. The chart's `setup-job` threading of `QUORUM_THRESHOLD` / `BLOCK_STALE_MEASURE` is for `AvsServiceManagerWrapper`, a different contract.

**No `operators` getter call sites.** Nothing in the workspace calls it, so the 5-field change is ABI-only here.

**Nothing consumes the horizon yet.** The router does not read the registry at all — it has no registry bindings outside `scripts/` — and `render_schnorr_payload` builds calldata for the user to submit rather than submitting itself, so there is no submission point to gate. Teaching the router to check `nextPossibleMutationBlock()` before rendering is a separate change, and only becomes meaningful once a deployment runs a non-zero window. Worth a follow-up issue.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings` (clean), `cargo check --all-targets --all-features`, and `cargo test --all-targets --all-features` — **252 tests pass**, including two new ones for the config parser.

The compile itself is the meaningful check on the binding regen: the 4-argument `SchnorrStakeRegistry::deploy` call only type-checks against a regenerated ABI, so a stale artifact would fail the build rather than fail at deploy time.

I have not run the full e2e (docker stack); the deploy path is exercised there and is worth a run before merge.


BreadrichEngels (commented):
## Review: thread the Schnorr notice window through deployment (service side of #69)

The code here is correct and minimal — I'd approve it on its own merits. I'm holding at COMMENT only because of merge coordination: this PR embeds bytecode regenerated from the **unmerged** solidity-sdk#69 branch, which I just left a REQUEST_CHANGES on (an announce-then-force double-add), and the paired e2e/Helm legs are still running.

### Code is correct ✅
- **`schnorr_notice_window()` / `parse_schnorr_notice_window()`** follow the established tolerant-parse convention (missing/empty/unparseable → default), and — correctly, unlike `max_queue_depth` — do **not** filter `0`, since `0` ("apply on commit, today's behavior") is a legitimate value. Tests cover both branches.
- **Correctly scoped, no lockstep hazard.** Unlike `STATE_ENCODING` (which must match on node+router or digests diverge, hence its fail-closed panic), `schnorr_notice_window` is read in exactly one place — the deploy script — and baked into the registry's immutable constructor. The node/router never read it, so a silent default can't cause consensus divergence. Good call not making this a lockstep var.
- **Deploy** passes `U256::from(notice_window)` as the 4th arg and logs it; matches #69's `(num, den, owner, noticeWindow)` constructor. Default `0` is genuinely right for the e2e (the set is force-registered before the target deploys via the immediate path, so no round is ever in flight and `noticeWindow` is inert there).
- **Helm-skip reasoning holds** — `helm/` has no `SCHNORR` reference; the registry is deployed by the local `deploy_array_summation` binary, so a chart value would be dead config. Consistent with what I noted on #332.

### Merge coordination — the actual gate
1. **The regenerated artifact is from the pre-fix #69 branch.** The *ABI shape* (4-arg constructor, 5-field `operators`) is stable regardless of the fix I requested (adding a `hasPendingChange` guard to the forced paths doesn't touch those signatures), so the compile-check that validates the 4-arg `deploy` call remains valid. But `scripts/bindings/abis/SchnorrStakeRegistry.json` also carries the **bytecode**, which *will* change when #69's double-add fix lands. So this artifact must be **re-regenerated from the final merged #69 commit** before it deploys anything real. Merging this before #69 is finalized would put pre-fix (buggy) registry bytecode into `main`'s deploy path. I'd invert the stated order: finalize #69 (with the fix) → regenerate this artifact from that commit → land both together.
2. **CI isn't green yet.** `Format/Clippy/Check` and both Docker builds pass, but the `e2e-test` legs (schnorr legacy + canonical/reentrant) and both Helm legs are still `in_progress` — and those are the only real exercise of the regenerated binding + 4-arg deploy (you flagged you hadn't run the full e2e). Please wait for them to go green before merge; a green schnorr e2e leg is what actually proves the new registry deploys and settles.

### Minor
- An unparseable `SCHNORR_NOTICE_WINDOW` silently falls back to `0`, which is the *least-protective* setting — a deployer who typos an intended non-zero window (`"3o0"`) silently gets no window. The deploy log (`notice window 0 blocks`) mitigates it, and it matches the file's tolerant-parse convention, but for a security-relevant knob consider `warn!`-ing (or rejecting) a non-empty-but-unparseable value rather than treating it identically to "unset."

Nothing to change in this diff itself — just don't merge ahead of a finalized #69 + a re-regenerated artifact + green e2e. Happy to give a clean approve once #69 settles and this artifact is rebuilt from it.

bagelface:
Re-regenerated the ABI artifact at `abc8684` — solidity-sdk#69 took another commit (`105e3c3`) adding `cancelChange(address)` so an emergency force no longer requires cancelling everything queued ahead of the target.

The ABI surface changed again: `cancelChange` added, and `hasPendingChange` (bool) became `pendingChangeIndex` (one-based uint). Neither is read anywhere in this workspace, so no call-site change was needed — but the artifact carries the deployed bytecode, so leaving it stale would have deployed the wrong registry.

Verified bytecode equality against a fresh build of `105e3c3`, and re-ran the gates: clippy `-D warnings` clean, full suite passing.

Merge order unchanged: this PR, then solidity-sdk#69.

PR #340, opened 2026-07-29
feat: end schnorr payload validity before the operator set can change

### LINKED ISSUE #339: Schnorr: router should honour the registry's operator-set change horizon
## Summary

`SchnorrStakeRegistry` gained a notice window in gas-killer/solidity-sdk#69 (design and rationale in gas-killer/solidity-sdk#60). Operator-set changes are announced ahead of time and `nextPossibleMutationBlock()` publishes the earliest block the set can change, so a round can check that the operator set it assembled against will still be the set that verifies its settlement.

**Nothing in this repo reads it.** The on-chain half is merged (#337 threads `SCHNORR_NOTICE_WINDOW` through deployment) but the horizon is currently inert.

## Why it is inert today, and when it stops being

`SCHNORR_NOTICE_WINDOW` defaults to `0` (`DEFAULT_SCHNORR_NOTICE_WINDOW`, `common/src/config.rs`), which applies changes as soon as they are committed. That is correct for the e2e stack — `deploy_schnorr` registers the whole operator set *before* the target contract deploys, so no round is ever in flight for a mutation to invalidate.

This becomes real the moment a deployment sets a non-zero window **and** expects the router to honour it. Until then the registry publishes a horizon nobody consults, so a mutation still invalidates in-flight rounds exactly as gas-killer/solidity-sdk#60 documents.

## Where the check belongs

Not at submission — **the router does not submit.** `render_schnorr_payload` (`router/src/executor.rs:858`) builds calldata, estimates gas and persists a user-signable payload; the on-chain submission is the user's. So there is no submission point to gate, and the check has to happen at render time.

## Suggested approach: clamp the existing payload expiry

A rendered payload already carries a `valid_until_block` (see `PAYLOAD_BLOCK_BUFFER` in `example.env`; `GET /tasks/{id}` returns a `PAYLOAD_EXPIRED` re-request error past it). That is exactly the right shape for this, so clamp it rather than adding a second notion of validity:

```
valid_until_block = min(valid_until_block, nextPossibleMutationBlock() - 1)
```

A payload then expires before the operator set can change, and a client that submits within its stated validity cannot eat an `InvalidQuorumSignature` caused by a set mutation. No new API surface, no new error path — an expired payload is already a re-request.

## What it costs

- **Registry bindings must reach the router.** `common/src/bindings/` holds only `gaskillersdk` and `schnorrgaskillersdk`; the registry binding lives in `scripts/bindings/schnorrstakeregistry.rs` and exists solely for the deploy binary, which the router cannot use. Add a registry binding under `common/src/bindings/` (the ABI artifact is already vendored and current as of #338).
- **One extra read per render.** `nextPossibleMutationBlock()` is a single storage load on-chain and can be fetched with the block number the submitter already queries (`schnorr_submitter.rs` calls `get_block_number` per attempt), so it need not add a round trip.
- Returns `type(uint256).max` when nothing is queued, so the clamp is a no-op in the common case — worth handling explicitly to avoid overflow when subtracting 1.

## Also: forced mutations bypass the horizon

`registerOperator` / `deregisterOperator` still mutate immediately, for bootstrap and emergencies such as a compromised key, and emit `ForcedMutation`. The horizon is a guarantee for the announced path only, with the registry's fail-closed watermark as the backstop. So a forced mutation — and any `OperatorRegistered` / `OperatorDeregistered` without a preceding `ChangeAnnounced` — must still be treated as "re-snapshot required"; the clamp does not make that case go away.

## Sizing note for whoever turns the window on

`noticeWindow` must exceed the signing round's duration plus `BLOCK_STALE_MEASURE`, since a rendered payload stays submittable for that long. The registry cannot enforce this: the staleness bound lives on each consumer contract and is adjustable there, while one registry may back several consumers. It is also **immutable** on the registry, so getting it wrong means a redeploy.

## Related

- gas-killer/solidity-sdk#60 — the problem, the design, and the "off-chain half" item this issue is
- gas-killer/solidity-sdk#69 — the registry-side implementation
- #337 — threaded `SCHNORR_NOTICE_WINDOW` through deployment (merged)
- #338 — refreshed the vendored ABI artifacts (merged)

bagelface:
Closes #339.

## Summary

`SchnorrStakeRegistry` publishes `nextPossibleMutationBlock()` — the earliest block its operator set can change (gas-killer/solidity-sdk#69, design in gas-killer/solidity-sdk#60). Nothing read it, so the horizon was inert and a mutation could still invalidate a payload the router had already handed out.

The router now reads it when rendering a Schnorr payload and ends that payload's validity before it:

```
valid_until_block = min(referenceBlock + PAYLOAD_BLOCK_BUFFER, horizon - 1)
```

A client that submits within the payload's stated validity therefore cannot eat an `InvalidQuorumSignature` caused by an announced operator-set change.

## Why this shape

**Not at submission — the router does not submit.** `render_schnorr_payload` builds calldata and persists a user-signable payload; the on-chain submission is the user's. So the check has to happen at render time.

**Reusing payload expiry rather than adding a mechanism.** A rendered payload already carries `valid_until_block`, and `GET /tasks/{id}` already answers a passed one with a `PAYLOAD_EXPIRED` re-request. Clamping that value expresses "this payload dies before the set can move" in machinery that exists — no new API surface, no new error variant, and an expired payload is already a re-request rather than a failure.

**Strictly below the horizon.** A settlement landing *on* the mutation block can already see the mutated set, hence `horizon - 1`.

**No new config.** `SchnorrGasKillerSDK.schnorrRegistry()` already exists, so the router discovers the registry from the target contract rather than needing an address threaded through the environment.

## Failure and edge behaviour

- **Nothing scheduled** — the registry returns `type(uint256).max`, so the clamp is a no-op. This is the common case, and the only case today since `SCHNORR_NOTICE_WINDOW` defaults to 0.
- **Horizon unreadable** (transient RPC failure, or the registry surface drifting) — logs a warning and leaves validity unclamped. A completed signing round should not be discarded for a transient error, and the registry's own fail-closed watermark still rejects a settlement assembled against a stale set. Same posture as the existing `estimate_gas` fallback immediately above it.
- **Change announced mid-round**, leaving the horizon at or behind the reference block — the payload is born expired and `GET /tasks/{id}` reports `PAYLOAD_EXPIRED`, so the client re-requests against the post-mutation set. Deliberate: preferable to handing over calldata that would revert. A zero horizon saturates rather than underflowing into `u64::MAX`, which would silently defeat the clamp.
- **Forced mutations still bypass it.** `registerOperator` / `deregisterOperator` skip the notice window and emit `ForcedMutation`; the horizon is a guarantee for the announced path only. Documented in the code and in `example.env` rather than implied.

## Registry binding: one function, pinned to the real ABI

The router needs exactly one view function, so `common/src/bindings/schnorrstakeregistry.rs` declares just that rather than vendoring `SchnorrStakeRegistry.json` a second time — `scripts` already carries the full artifact because it deploys the registry, and #338 was a reminder that duplicated artifacts drift.

The obvious objection is that a hand-written declaration can drift silently. Two things prevent that:

- The router's own call site pins it at **compile time** — I verified a rename fails the build with `no method named nextPossibleMutationBlock`.
- `router_registry_declaration_matches_the_deploy_artifact` (in `scripts/bindings`, which owns the artifact) asserts the hand-written selector equals the artifact-generated one, so an **upstream** rename fails a test instead of degrading into a runtime warning. Verified non-vacuous: with the declaration renamed and the call site fixed, the test fails on `left: [246, 119, 49, 145] / right: [109, 17, 169, 122]`.

## Tests

Five cases on `clamp_to_mutation_horizon`, which is a free function so the arithmetic is testable without mocking a provider: no-op on `U256::MAX`, no-op on `None`, no-op when the horizon is beyond validity, the `horizon - 1` boundary (including horizon == validity), and the born-expired case with the zero-horizon underflow guard. Plus the selector pin above.

**258 tests pass** (252 before). `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings` and `cargo check --all-targets --all-features` all clean.

## Scope

**#339 only.** gas-killer/solidity-sdk#66 changes the Schnorr `verifyAndUpdate` signature (drops `targetFunction`, adds `anchorHash` / `callerAddress` / `contractCalldata`), and the service side of that is unstarted. It is near-orthogonal to this change — #66 churns calldata construction and the signed digest, this touches `valid_until_block` in the same function — so the rebase is cheap and this does not front-run an unmerged PR.

No behaviour change for any current deployment: every stack runs `SCHNORR_NOTICE_WINDOW=0`, where the horizon is unbounded and the clamp does nothing. The two extra reads per Schnorr render are the only observable difference until a window is set.

### DIFF
--- a/common/src/bindings/mod.rs
+++ b/common/src/bindings/mod.rs
@@ -33,3 +33,14 @@ pub mod gaskillersdk;
     dead_code
 )]
 pub mod schnorrgaskillersdk;
+
+#[allow(
+    non_camel_case_types,
+    non_snake_case,
+    clippy::pub_underscore_fields,
+    clippy::style,
+    clippy::empty_structs_with_brackets,
+    missing_docs,
+    dead_code
+)]
+pub mod schnorrstakeregistry;
--- a/common/src/bindings/schnorrstakeregistry.rs
+++ b/common/src/bindings/schnorrstakeregistry.rs
@@ -1,0 +1,16 @@
+use alloy::sol;
+
+// The router reads a single view function off `SchnorrStakeRegistry`, so it declares just that
+// function rather than vendoring the contract's ABI artifact a second time (the deploy tooling in
+// `scripts` already carries the full artifact, bytecode included, because it deploys the registry).
+// `scripts::bindings` pins this declaration against that artifact so an upstream rename fails a
+// test rather than silently degrading into a warning at runtime.
+sol! {
+    #[sol(rpc)]
+    interface ISchnorrStakeRegistry {
+        /// The earliest block at which the operator set can change, or `type(uint256).max` when no
+        /// change is scheduled. A settlement is safe from set-mutation invalidation while the block
+        /// it lands in is below this value.
+        function nextPossibleMutationBlock() external view returns (uint256);
+    }
+}
--- a/example.env
+++ b/example.env
@@ -178,11 +178,16 @@ FORK_URL=https://ethereum-sepolia-rpc.publicnode.com
 #
 # SCHNORR_NOTICE_WINDOW: blocks a SchnorrStakeRegistry operator-set change must be
 # announced ahead of taking effect (schnorr mode only; see SIGNATURE_SCHEME). Announcing
-# publishes a horizon via nextPossibleMutationBlock() that a round can check, so a
-# settlement is never assembled against one operator set and verified against another.
+# publishes a horizon via nextPossibleMutationBlock(); the router reads it when rendering a
+# payload and ends that payload's validity before it, so a settlement is never assembled
+# against one operator set and verified against another.
 # Defaults to 0, which applies changes as soon as they are committed — correct for this
 # stack, where the whole operator set is registered before the target contract deploys and
 # no round is ever in flight for a mutation to invalidate.
+#
+# The horizon only covers announced changes. registerOperator / deregisterOperator bypass
+# the window for bootstrap and emergencies and emit ForcedMutation; a forced change can
+# still invalidate a payload the router already handed out.
 #
 # A deployment that changes its operator set while rounds are in flight must set this
 # above the signing round's duration plus BLOCK_STALE_MEASURE, since a rendered payload
@@ -199,6 +204,10 @@ BLOCK_STALE_MEASURE=300
 # returns a PAYLOAD_EXPIRED re-request error instead of calldata that would revert on-chain. A
 # payload is only submittable while referenceBlockNumber + BLOCK_STALE_MEASURE >= block.number, so
 # values above BLOCK_STALE_MEASURE are clamped to it. Default: 50.
+#
+# In schnorr mode this is an upper bound rather than the final value: a payload's validity is also
+# clamped to end before the registry's next possible operator-set change (see
+# SCHNORR_NOTICE_WINDOW), so it can expire sooner than this buffer allows.
 # PAYLOAD_BLOCK_BUFFER=50
 #
 # The GET /tasks/{id} block-window freshness check measures valid_until_block against the L1 chain
--- a/router/src/executor.rs
+++ b/router/src/executor.rs
@@ -15,6 +15,7 @@ use gas_killer_common::bindings::gaskillersdk::{
     BN254, GasKillerSDK, IBLSSignatureCheckerTypes as GasKillerIBLSTypes,
 };
 use gas_killer_common::bindings::schnorrgaskillersdk::SchnorrGasKillerSDK;
+use gas_killer_common::bindings::schnorrstakeregistry::ISchnorrStakeRegistry;
 use gas_killer_common::bindings::{GAS_KILLER_INTERFACE_ID, SCHNORR_GAS_KILLER_INTERFACE_ID};
 use gas_killer_common::{BundleProof, PayloadView, TaskBundle};
 use std::collections::HashMap;
@@ -126,6 +127,27 @@ struct PreparedSchnorr<P> {
 struct RenderedRound {
     payload: PayloadView,
     bundle: TaskBundle,
+}
+
+/// Bound a rendered payload's validity so it expires before the operator set can change.
+///
+/// A payload is submittable until `valid_until_block`, but a Schnorr settlement only verifies while
+/// the operator set still matches the one its signature was assembled against. `horizon` is the
+/// registry's earliest possible mutation block, so validity must end strictly below it — a
+/// settlement landing *on* that block can already see the mutated set.
+///
+/// `None` leaves the value untouched (the horizon could not be read). `u64::MAX` — which the
+/// registry returns as `type(uint256).max` when nothing is scheduled — is the common case and
+/// clamps nothing. A horizon at or below the existing validity yields a payload that is already
+/// expired; that is deliberate, and `GET /tasks/{id}` answers it with the usual `PAYLOAD_EXPIRED`
+/// re-request rather than handing over calldata that would revert.
+fn clamp_to_mutation_horizon(valid_until_block: u64, horizon: Option<U256>) -> u64 {
+    let Some(horizon) = horizon else {
+        return valid_until_block;
+    };
+    // Saturating: an unscheduled horizon is `type(uint256).max`, far beyond u64.
+    let horizon = u64::try_from(horizon).unwrap_or(u64::MAX);
+    valid_until_block.min(horizon.saturating_sub(1))
 }
 
 /// Handler for executing verifyAndUpdate transactions with multi-chain support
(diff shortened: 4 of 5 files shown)


BreadrichEngels (approved):
## Review: end Schnorr payload validity before the operator set can change (closes #339)

**Verdict: Approve.** This closes exactly the "nothing consumes `nextPossibleMutationBlock()` yet" gap I flagged on #332/#337 — the horizon the notice-window work built (solidity-sdk #69) is now actually read at render time. The clamp arithmetic is correct against every edge, it's fail-safe (it can only make a payload expire *earlier*, never admit a bad settlement), and all 9 checks are green including all three e2e legs. No audit gate here: this touches payload validity, not the verification/consensus path.

### The clamp is correct
`valid_until_block = min(reference + buffer, horizon - 1)`, and I walked every branch of `clamp_to_mutation_horizon`:
- **`None`** (horizon unreadable) → unclamped. Correct fail-open, backstopped by the registry's own watermark.
- **`U256::MAX`** (nothing scheduled) → `try_from` errors → `u64::MAX`, `saturating_sub(1)` → `MAX-1`, `min` → no-op. The common case, and the default since `SCHNORR_NOTICE_WINDOW=0`.
- **strictly below the horizon** — `horizon 150, valid 150 → 149`; `horizon 151 → 150` (unchanged); `horizon 120 → 119`. The "settlement landing *on* the mutation block already sees the mutated set" reasoning is right, and the `== validity` boundary is handled.
- **born-expired** (`horizon 10 → 9`) → `GET /tasks/{id}` reports `PAYLOAD_EXPIRED`, a re-request rather than revert-bound calldata.
- **zero horizon** → `saturating_sub(1) = 0`, not an underflow to `u64::MAX`. This is the one that would silently defeat the whole clamp if done with a bare `- 1`, and it's both guarded and explicitly tested. Good.

The direction is sound because the horizon can only move *later* between render and submit on the announced path (FIFO `eligibleBlock` is non-decreasing, cancel only pushes it later, a new announcement appends) — so a payload valid until `horizon-1` lands before any *announced* set change. Reading it from `schnorrRegistry()` off the target (rather than threading an address through env) is the right call.

### Scope and fail-modes are handled correctly
- **Forced mutations bypass it** — correctly out of scope and documented in both the code and `example.env`; the registry's fail-closed watermark remains the backstop, so the worst case is a chain rejection, never a bad settlement.
- **Horizon-read failure** → warn + unclamped, matching the `estimate_gas` fallback posture. A completed round shouldn't be discarded for a transient RPC blip, and the chain still fail-closes. It's a UX degradation (user eats a rejection instead of a clean `PAYLOAD_EXPIRED`), not a safety hole.
- **No behavior change for current deployments** — `SCHNORR_NOTICE_WINDOW=0` ⇒ unbounded horizon ⇒ no-op, confirmed by the green schnorr e2e legs.

### Binding-drift guard is a nice touch
Declaring just the one view function by hand (rather than re-vendoring the full artifact, which #338 showed drifts) is the right trade-off, *because* you pinned it: the call site pins the selector at compile time, and `router_registry_declaration_matches_the_deploy_artifact` asserts the hand-written selector equals the artifact-generated one, so an upstream rename fails a test instead of degrading to a runtime warning. Verified-non-vacuous, per the writeup.

### Non-blocking notes
1. **The selector pin covers the signature, not the return type.** `…Call::SELECTOR` is derived from the function signature only, so a hypothetical change to `nextPossibleMutationBlock`'s *return* type (uint256 → something else) would keep the selector equal and slip past the test. It degrades gracefully — alloy's decode would error → `None` → unclamped (fail-open) — so it's low risk (view return types are stable), but worth a comment noting the pin is selector-only.
2. **Two extra reads per Schnorr render.** The registry address from `schnorrRegistry()` is effectively stable per target, so caching it (keyed by target) would drop one of the two reads; only `nextPossibleMutationBlock()` genuinely needs re-reading. Minor, render-path-only, not blocking.

Clean, well-tested, and it's the piece that makes the whole notice-window chain (sdk #60/#69 → service #337 → this) actually do something. Nice.

BreadrichEngels (approved):
## Follow-up review at `f0e0598` — approval stands

The only change since my approval (`2a7061f`) is the merge commit `f0e0598` bringing in current `main` (post-#349 `src/` layout, post-#353 unbounded profile). Everything below it in the branch history is mainline, not new work on this PR. The feature diff is unchanged in shape (+176/−3 across the same 5 files), and per the PR notes `router/src/executor.rs` and `common/src/bindings/` were untouched by `main` — so the clamp itself (`clamp_to_mutation_horizon`, `valid_until_block = min(reference + buffer, horizon - 1)`) and its five edge tests are exactly what I already verified. No re-analysis of the arithmetic needed.

I did verify the one thing the merge actually moved — the cross-crate selector drift-guard. `router_registry_declaration_matches_the_deploy_artifact` now lives at `scripts/src/bindings/mod.rs` (was `scripts/bindings/mod.rs` before #349 deleted it) and still asserts the router's hand-written `gas_killer_common::bindings::…::nextPossibleMutationBlockCall::SELECTOR` equals the deploy artifact's `crate::bindings::…` one. The conflict was resolved correctly: the guard is intact at the new path, so an upstream rename still fails a test rather than degrading to a runtime warning + unclamped validity.

Everything from the original approval carries over unchanged — correct-and-fail-safe clamp (can only expire a payload *earlier*, never admit a bad settlement), no behaviour change for current deployments (`SCHNORR_NOTICE_WINDOW=0` ⇒ `type(uint256).max` horizon ⇒ no-op), and the two non-blocking notes (selector-pin covers signature not return type; the `schnorrRegistry()` address is cacheable per target). CI is re-running after the merge — the substance here is orthogonal to what `main` moved (the horizon touches `valid_until_block`; #353 added the unbounded profile leg), so I'd expect it green, and would flag it if a leg unexpectedly fails. Approving.

bagelface:
Both non-blocking notes addressed in `d0a473c` and `116b0ec`.

**1. Selector pin covers the signature, not the return type** — fixed rather than commented. The test now round-trips `U256::MAX` through both declarations: the artifact's `abi_encode_returns` and the router's `abi_decode_returns` have to agree on the type, so the pin covers the return as well as the signature.

Two things came out of verifying it, both slightly different from the note:

- The failure mode was worse than "decode would error → `None` → unclamped". `uint128` and `uint256` are both single 32-byte words, so a narrowing upstream would have decoded *silently and successfully* into whatever the hand-written declaration claimed. Not fail-open — just wrong, with no signal. Worth closing for that reason rather than the low-risk one.
- Part of it was already covered, but not by the test. Narrowing the return in the router's own declaration fails the **router lib** build, because `schnorr_mutation_horizon` returns `Option<U256>` into `clamp_to_mutation_horizon` — the same call-site pin you credited for the signature. What was genuinely unguarded is the upstream direction: artifact moves, hand-written declaration doesn't. That's what the round-trip catches, verified non-vacuous by narrowing the artifact's `outputs[0]` to `uint128` — `expected &u128, found &Uint<256, 4>`, failing in `scripts` lib test.

**2. Caching the registry address** — I looked at making it a per-target memo alongside `interface_cache` and decided against it; the comment now records why.

`schnorrRegistry()` isn't immutable. `SchnorrGasKillerSDK` keeps it in ERC-7201 storage behind `_setSchnorrRegistry`, an **`internal`** setter — both in-repo examples call it only from their constructor, but nothing stops an integrator target from exposing an owner-gated path that re-points it. That's the difference from `interface_cache`: a deployed contract's ERC-165 support genuinely cannot change, so that memo can never go stale.

A cached address would outlive a re-point for the lifetime of the process and read the horizon off a registry the target no longer uses. Both outcomes are bad and quiet — the abandoned registry almost certainly has nothing scheduled, so it returns `type(uint256).max` and the clamp no-ops exactly when a real change is pending on the new one.

Against that, the saving is one `eth_call` per *completed round*, not per poll — rendering is not on a hot path, as you noted. Trading a silent-staleness hazard on the value the whole clamp keys off for that seemed like the wrong side of the trade. Happy to revisit if you read the mutability differently.

PR #345, opened 2026-08-04
feat: sign the slashable message (anchorHash, caller, calldata)

bagelface:
## Summary

Re-ports #310 onto `main`. That PR was stacked on the closed ECDSA line (#309) and carried a
vendored `contracts/` tree this repo no longer has, so it could not be rebased — the Rust changes
are re-applied here against current `main`, and the ECDSA work is dropped.

Migrates the operator signing path to the **slashable message format**, so incorrect storage
updates signed by a quorum can be proven wrong and slashed on-chain:

```
OLD  sha256(abi.encode(transitionIndex, address(this), targetFunction,  storageUpdates))
NEW  sha256(abi.encode(transitionIndex, address(this), anchorHash, callerAddress, contractCalldata, storageUpdates))
```

The 4-byte selector is replaced by the full calldata, plus the caller and the anchor block hash —
everything a challenger needs to reproduce the exact execution.

**Both schemes.** `main` has gained Schnorr since #310, and solidity-sdk#66 moves both SDKs to the
same six-field digest, so this covers the BLS and Schnorr paths alike.

## Depends on

gas-killer/solidity-sdk#66 — merged onto SDK `main` first (it had drifted 21 commits behind:
payable `verifyAndUpdate`, `TransitionGuard`, `SchnorrStakeRegistry` notice window). The vendored
artifacts here are built from that merged tree. **This must not merge before #66.**

## Changes

- **`common/src/task_data.rs`**: `GasKillerTaskData` gains `anchor_hash: B256`;
  `build_payload_hash` ABI-encodes the six-field tuple with alloy `abi_encode_params`. A
  golden-vector test pins the digest to an independent `cast abi-encode | sha256` value, proving
  byte-parity with Solidity `abi.encode` without needing a chain. `function_selector()` is gone —
  nothing signs or submits a bare selector any more.
- **`common/src/validator.rs`**: derives `anchor_hash` from the **same** EVMSketch executor that
  produced the storage updates (a cache hit — no extra RPC), so it is always the hash of the block
  the updates were computed at. `expected_digest_for_task` **rejects** a task whose announced
  anchor differs from the derived one, so a malicious task creator cannot get honest operators to
  sign a commitment no challenger could reproduce. `anchor_hash` also joins `DigestCacheKey` —
  without it a forged-anchor task could hit an honest entry and skip that check entirely.
- **`common/src/payload.rs`**: `TaskBundle` swaps `target_function` for `anchor_hash`,
  `caller_address`, `contract_calldata`, keeping its "every `verifyAndUpdate` argument component"
  invariant true.
- **`router/`**: `anchor_hash` threads from `enrich()` through `EnrichedTask` into the task data;
  `PreparedBls`/`PreparedSchnorr` and all four `verifyAndUpdate` builders (broadcast + render, per
  scheme) submit the new arguments. The structs already carried `from_address`, which is
  `callerAddress` by construction, so it is reused rather than duplicated.
- **Vendored ABIs + ERC-165 ids**: regenerated from the merged SDK.
  `GAS_KILLER_INTERFACE_ID` `0x93de4531` → `0x2a79d7ba`, `SCHNORR_GAS_KILLER_INTERFACE_ID`
  `0x82b35a01` → `0xb95d5f32`. A new unit test asserts each constant equals the corresponding
  binding's `verifyAndUpdateCall::SELECTOR`, so a future ABI refresh that moves a selector fails in
  `cargo test` rather than in a live run (the router's preflight hard-fails every submission on a
  mismatch).
- **`scripts/verify_message_hash_parity.rs`**: vectors carry the execution context and sweep
  **both** dynamic tails across the ABI padding boundaries. Both SDKs expose the same
  `getMessageHash(uint256,bytes32,address,bytes,bytes)` selector, so the check stays
  scheme-agnostic.
- **BLS e2e self-deploys its factory**: the BLS demo target was deployed through a factory
  pre-deployed at a fixed address by the `ghcr.io/breadchaincoop/eigenlayer` container, which
  embeds the old digest — BLS e2e would fail against the new format. `ArraySummationFactory` is now
  vendored with bytecode and `deploy_array_summation` deploys its own, mirroring what the Schnorr
  and reentrant paths already do. `ARRAY_SUMMATION_FACTORY_ADDRESS` remains as an override for
  deployments that reuse a long-lived factory.

## Validation

- `cargo fmt --all`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo check`,
  `cargo test --all-targets --all-features`: **254 tests pass, 0 failures**.
- solidity-sdk on the merged #66 branch: `forge fmt --check`, `forge build --sizes`, `forge test` —
  **189 tests pass, 0 failures**.
- The e2e's step 7b (`verify_message_hash_parity`) is the real end-to-end gate: it calls the
  deployed contract's `getMessageHash` and fails the run on any divergence.

## Notes for reviewers / deployers

- **Coordinated breaking change.** The digest and the `verifyAndUpdate` ABI both change. Deployed
  targets built from the old SDK will reject these submissions, and the ERC-165 preflight will
  refuse them first. Targets must be redeployed from the merged SDK. Everything is pre-production,
  so there is no on-chain migration.
- **Helm** `deployTarget` builds targets from the `ghcr.io/gas-killer/solidity-sdk` image; that tag
  needs to point at a build of the merged SDK. Deploy-time knob, no code change.
- **Persisted bundles.** `TaskBundle` is stored as JSON in SQLite. Rows written before this change
  no longer deserialize and surface as a 500 from the payload read path — acceptable
  pre-production, but worth knowing if a dev database is carried across.

## Out of scope (follow-ups)

Slasher deployment and EigenLayer operator-set enrollment are deliberately not here. `_setSlasher`
is `internal` in the SDK, so no target can configure a slasher yet, and operators hold no slashable
magnitude — a slash would be a no-op on-chain. That needs operator-set creation, magnitude
allocation, and a PermissionController grant, plus product decisions on strategies and delays. The
concrete Schnorr slasher is solidity-sdk#65.

Closes #310.


bagelface:
## CI: helm yield-distribution leg disabled (#346)

First CI run was green everywhere except `Test with Helm Charts (bls)`, which failed only at
**"Trigger Yield Distribution via router"**. Worth being precise about what did and did not fail,
since the earlier steps are the ones that validate this PR:

| step | result |
|---|---|
| Deploy ArraySummation contract | ✅ (self-deployed factory works in-cluster) |
| Trigger Gas Killer multiple times and verify execution | ✅ (full BLS rounds on the new digest) |
| Trigger Yield Distribution via router | ❌ |

```
INFO  Message hash match confirmed
WARN  Target contract does not support GasKiller interface  interface_id=0x2a79d7ba
```

The digest is right; the ERC-165 preflight is refusing the target. That target is the
YieldDistributor **already deployed on Gnosis** (`0x37A9CC…06da`), reached by forking Gnosis and
impersonating the owner — there is no deploy step, so the job runs live bytecode that predates this
change and still advertises `0x93de4531`. The preflight is doing its job: it stopped a submission
that would have reverted.

That contract can't be rebuilt from this repo, so the leg is disabled here (helm-install flag plus
the trigger step, both commented and pointing at #346) and #346 tracks redeploying it — preferably
by deploying a fresh one onto the fork from a vendored artifact, the same treatment
`ArraySummationFactory` got in this PR, so a future coordinated ABI change doesn't break it the same
way.

PR #347, opened 2026-08-06
feat: manifest-driven harness for deploying and targeting example contracts

bagelface:
## Why

`gas-killer/example-contracts` is meant to be a public library showing off Gas Killer use cases, but there was no way to get one of those contracts onto a chain and pointed at by the AVS without writing Rust. `deploy_array_summation` is hardcoded to a single target deployed through a pre-existing factory, and `scripts/bindings/` needs a compiled-in `sol!` module plus a vendored artifact per contract.

This adds a harness that fetches and builds the example contracts, deploys one wired to the AVS, records its address where the existing tooling already looks, and emits a runnable scenario — with no new Rust needed for the next example someone adds.

## What

- **`scripts/deploy_example.rs`** — manifest-driven deployer. Constructor argument *types* come from the Foundry artifact's ABI at runtime (`alloy-dyn-abi`) and *values* from TOML, so adding an example is a manifest edit. Deploys by appending encoded args to the creation bytecode, runs post-deploy assertions, executes setup calls as any declared signer, records `addresses.<name>`, and writes a scenario file.
- **`scripts/examples/`** — `fetch_examples.sh` (clone + recursive submodules + `forge build` at a pinned SHA), `examples.toml` (both examples), `run_example.sh` (fetch → deploy → run scenario).
- **`submit` in `run_scenario`** — see below. Defaults to `false`, so every existing scenario file behaves exactly as before.
- **`scripts/bindings/task_payload.rs`** — the poll/submit helpers, previously private to `send_request`, now shared by both bins.

Nothing existing changes behaviour: `deploy_array_summation`, `run_e2e_test.sh`, and the `array-summation`/`reentrant` CI matrix are untouched.

### The `submit` gap

With `INGRESS=true` the router *renders* a `verifyAndUpdate` transaction rather than broadcasting it, so nothing reaches the chain until the caller signs and sends it. `run_scenario` posted the task and then polled `stateTransitionCount()` — which no one was going to move, so `verify = true` could only ever time out. `submit = true` closes that loop.

### Why not reuse the examples repo's own forge scripts

`script/DeployBase.sol`'s `_resolveChecker()` deploys a **`MockBLSSignatureChecker`** when `SIG_CHECKER_ADDRESS` is unset, and `_avs()` defaults to `address(0xA75)` — against the real AVS that yields a target which accepts unsigned diffs. Separately, wiring a target to `addresses.blsSigCheck` (the `BLSSigCheckOperatorStateRetriever`) makes `verifyAndUpdate` revert with an empty `0x`. The deployer resolves the checker by an explicit precedence, rejects the retriever by name, and probes for `registryCoordinator()` before spending gas.

## Verified against a live local stack

`guardedVault` runs green end to end on an anvil Sepolia fork:

```
🚀 deployed at 0x32b9698f… (tx 0x5c6fd5a2…)
✅ supportsInterface(0x93de4531) = true, stateTransitionCount() = 0
⚙️  3 setup call(s): deposit(uint256) × signers 1, 2, 3
✅ task ready: estimated_gas=288363 valid_until_block=11425518
✅ verifyAndUpdate landed: tx 0xd22449bf… in block 11425470
stateTransitionCount: 1 (initial: 0)
━━━  Overall: 1/1 passed
```

On-chain post-state confirms a real, invariant-preserving transition: `totalShares` still 3000 (conservation held), shares 1000/1000/1000 → **1100/900/1000**, `isHealthy() = true`. The `Settled(uint256)` LOG1 event was replayed byte-exact by the diff under the default `STATE_ENCODING=legacy`.

Also green: `cargo fmt`, `clippy --all-targets --all-features -D warnings`, `cargo test` (291 tests, 41 new).

## `onchainLife` requires the `prestate-net` encoding

`onchainLife` deploys and passes the routability gate under any encoding, but only **settles** under `STATE_ENCODING=prestate-net`.

`legacy` and `canonical` derive the diff from a struct-log `debug_traceCall`, costing `O(execution steps)`, and one generation of Life is ~16.5M gas (measured: 16,552,726). On a stock anvil that trace drives the process from 8 MB to 2.2 GB of RSS in 20s and still has not returned; inside the 3.82 GiB compose container it is OOM-killed outright (`exit 137`), so the task never reaches `ready`:

```
ERROR gas_killer_router::sequencer: failed to enrich task, dropping request
  error=Gas analysis failed: debug_trace_call failed: error sending request
```

`generations` is already at its floor of 1, so no manifest setting avoids this — the limit is the encoder. Under `prestate-net` the same call returns in **0s with a 3-slot diff**, and `callTracer` shows no target-depth frames, so it takes the net form rather than the canonical fallback.

That encoding landed in #348 (merged, `51a61f2`), so it is available on `main` and this branch is rebased onto it. Note the value is fail-closed by design: `parse_state_encoding` panics on anything unrecognized rather than falling back, so a typo cannot put an operator into a running-but-digest-divergent state.

**Verified end to end** on a rebuilt local stack with all four binaries on `prestate-net`: `generation()` 0 → 1, `stateTransitionCount()` 0 → 1, `boardHash()` changed, the `GenerationStepped` LOG2 event replayed by the diff, and anvil memory flat at ~30 MiB throughout (see the verification comment below). `run_example.sh` warns when `onchainLife` is run under any other encoding, and the manifest entry is kept so the case stays one command:

```bash
STATE_ENCODING=prestate-net docker compose up -d --force-recreate
./scripts/examples/run_example.sh onchainLife
```

<sub>An earlier version of this description claimed gas-analyzer#163 needed "no public API change / zero code changes" on the service side. That came from #163's own body, which is stale: a commit inside that PR (`60ab3f44`) deliberately turned the implicit fast path into the opt-in `StateEncoding::PrestateNet` variant and removed the graceful fallback. Corrected above.</sub>

## Follow-up

`router/Cargo.toml` and `scripts/Cargo.toml` both declare `gas-analyzer` but never use it; `common` is the only real consumer. Unrelated cleanup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,11 @@ test_processes.pids
 # Script build artifacts
 scripts/target
 
+# Example contracts checkout (fetched by scripts/examples/fetch_examples.sh) and the
+# scenario files deploy_example generates from it.
+/.examples
+scripts/scenarios/generated/
+
 # Helm chart dependencies (fetch via `helm dependency update`)
 helm/gas-killer/charts/*.tgz
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2213,6 +2213,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b193af5b67834b676abd72466a96c1024e6a6ad978a1f484bd90b85c94041351"
 dependencies = [
  "clap_builder",
+ "clap_derive",
 ]
 
 [[package]]
@@ -2225,6 +2226,18 @@ dependencies = [
  "anstyle",
  "clap_lex",
  "strsim",
+]
+
+[[package]]
+name = "clap_derive"
+version = "4.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1110bd8a634a1ab8cb04345d8d878267d57c3cf1b38d91b71af6686408bbca6a"
+dependencies = [
+ "heck",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -8174,9 +8187,12 @@ name = "scripts"
 version = "0.1.0"
 dependencies = [
  "alloy",
+ "alloy-dyn-abi",
+ "alloy-json-abi",
  "ark-bn254",
  "ark-ec",
  "ark-ff 0.5.0",
+ "clap",
  "dotenv",
  "gas-analyzer-evmsketch",
  "gas-killer-common",
@@ -8185,6 +8201,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
+ "tempfile",
  "tokio",
  "toml 0.8.23",
  "url",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,9 @@ license = "PPL"
 
 [workspace.dependencies]
 alloy = "1.0"
+alloy-dyn-abi = "1.0"
 alloy-eips = "1.0"
+alloy-json-abi = "1.0"
 alloy-network = "1.0"
 alloy-primitives = "1.0"
 alloy-provider = "1.0"
--- a/example.env
+++ b/example.env
@@ -379,3 +379,28 @@ ARRAY_SUMMATION_MAX_VALUE=1000
 ARRAY_SUMMATION_SEED=42
 ARRAY_SUMMATION_FACTORY_ADDRESS=0xF7ded769418Ec1Db4DA3bd2d47ab72ce2296A032
 # ARRAY_SUMMATION_ADDRESS=0x8867cf9D1FfC33fE4761c19E621Ee29323E91E56 # Setting this will override deploying a new contract
+
+# =============================================================================
+# Example Contracts Harness (scripts/examples/)
+# =============================================================================
+# The public example-contracts library is fetched and built by
+# scripts/examples/fetch_examples.sh, then deployed by the `deploy_example` bin, which
+# reads the manifest at scripts/examples/examples.toml. See scripts/README.md.
+#
+# Where the checkout and its forge artifacts live. Gitignored.
+# EXAMPLES_DIR=.examples/example-contracts
+# Git revision to build. Pinned to a SHA so a rebuild is reproducible; override with a
+# branch name to track upstream.
+# EXAMPLES_REF=f9aa0a6020ec7b3e28db47999ece8ba54b3847f6
+# EXAMPLES_REPO=https://github.com/gas-killer/example-contracts
+#
+# Constructor wiring for deployed examples. Both default to reading the AVS deployment
+# JSON at AVS_DEPLOYMENT_PATH, which is what the local compose stack produces; set them
+# explicitly when deploying against a testnet with no local deployment JSON.
+# EXAMPLE_AVS_ADDRESS=          # default: addresses.avsServiceManagerWrapper
+# EXAMPLE_SIG_CHECKER_ADDRESS=  # default: addresses.IncredibleSquaringTaskManager
+#
+# NOTE: the signature checker must be a real BLSSignatureChecker (one exposing
+# registryCoordinator()). addresses.blsSigCheck is the BLSSigCheckOperatorStateRetriever,
+# which has no checkSignatures — a target wired to it reverts with an empty 0x during
+# verifyAndUpdate. deploy_example rejects it rather than letting that happen silently.
--- a/example.env
+++ b/example.env
@@ -387,6 +387,11 @@ ARRAY_SUMMATION_FACTORY_ADDRESS=0xF7ded769418Ec1Db4DA3bd2d47ab72ce2296A032
 # scripts/examples/fetch_examples.sh, then deployed by the `deploy_example` bin, which
 # reads the manifest at scripts/examples/examples.toml. See scripts/README.md.
 #
+# Per-example encoding requirement: guardedVault settles under any STATE_ENCODING, but
+# onchainLife only settles under prestate-net (its tracked function is ~16.5M gas, which
+# the struct-log encoders behind legacy/canonical cannot extract). See the STATE_ENCODING
+# block above.
+#
 # Where the checkout and its forge artifacts live. Gitignored.
 # EXAMPLES_DIR=.examples/example-contracts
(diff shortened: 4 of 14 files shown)


bagelface:
### `onchainLife` verified end to end under `prestate-net`

Now that #348 has landed on `main`, this branch is rebased onto it and the one claim that was never provable is proven. Full stack rebuilt so all four binaries carry the encoding — checked rather than assumed, since a stale node image would fail quorum:

```
node-1: prestate-net   node-2: prestate-net   node-3: prestate-net   router: prestate-net
```

Deploy and settle:

```
🚀 deployed at 0x83068895… (tx 0x371f4e62…)
✅ supportsInterface(0x93de4531) = true, stateTransitionCount() = 0
✅ task ready: estimated_gas=309989 valid_until_block=11459653
Submitting verifyAndUpdate … (1956 bytes calldata)
✅ verifyAndUpdate landed: tx 0x9c9bbc94… in block 11459605
stateTransitionCount: 1 (initial: 0)
━━━  Overall: 1/1 passed
```

On-chain post-state — the board genuinely advanced one generation, it isn't just a counter bump:

| | before | after |
|---|---|---|
| `generation()` | 0 | **1** |
| `stateTransitionCount()` | 0 | **1** |
| `boardHash()` | `0xcd4c41b2…` | **`0x9604934a…`** |

The `GenerationStepped(uint256,bytes32)` LOG2 event was replayed by the diff with `topic0 = 0x617d56c3…` (matches `cast keccak`) and indexed generation `1`.

**Anvil memory stayed flat: 29.58 MiB → 30.76 MiB across the whole round.** That is the result, not a footnote — it is also what proves the prestate path actually ran rather than the canonical fallback. The struct-log encoder on this exact call drove anvil from 8 MB to 2.2 GB in 20s without returning, and OOM-killed the 3.82 GiB container (`exit 137`). Same call, same contract, ~1 MiB of growth instead.

Ready for review.

BreadrichEngels (approved):
## Review: manifest-driven example-contract deploy harness

**Approve.** This is dev/test tooling (all under `scripts/`), not node/router runtime — and the one genuinely safety-relevant thing a deploy harness can get wrong (wiring a target to a checker that doesn't actually verify signatures → a target that accepts *unsigned* diffs) is handled carefully. Existing behaviour is untouched, `submit` is opt-in and defaults off, and the whole thing is verified live and green across all 10 checks including the new `prestate-net` leg. I focused on the checker-resolution and submit paths and trusted the dyn-abi/TOML/assertion machinery given the 291 tests + the end-to-end run on the PR.

### The checker-wiring safety is the right thing to have gotten right
`validate_sig_checker` requires the resolved checker to (a) have code and (b) answer `registryCoordinator()` — which a real `BLSSignatureChecker` does and the `BLSSigCheckOperatorStateRetriever` doesn't — and `reject_retriever_as_checker` independently fails fast if the resolved address equals `addresses.blsSigCheck`. Defaulting the checker to `IncredibleSquaringTaskManager` (which inherits `BLSSignatureChecker`) rather than the retriever, and resolving wiring explicitly instead of going through the examples repo's `DeployBase._resolveChecker()`, is what sidesteps the two documented footguns: the `MockBLSSignatureChecker` that would accept unsigned diffs against the real AVS, and the retriever that makes `verifyAndUpdate` revert with empty `0x`. Good — those are exactly the failures you'd want a harness to refuse rather than silently deploy.

### submit closes a real gap, safely
The `submit = true` path correctly fixes the "poll `stateTransitionCount()` that no one will move" dead end under `INGRESS=true`: it signs the rendered payload with `FUNDED_KEY`/`PRIVATE_KEY` (env-provided, errors if absent) and broadcasts, ordered *before* the verify poll so the transition it watches for is the one this submission causes. It's gated on `submit && api_success && status == 202`, requires `http_rpc`, and — critically — defaults to `false` with a test pinning that existing scenario files behave exactly as before. The `send_request` → `task_payload.rs` extraction is a behaviour-preserving refactor (the `-161/+15` on `send_request` is mostly code moved, now shared by both bins).

### One scope caveat (not a change request)
The checker validation proves "not the retriever, and has `registryCoordinator()`" — not "this checker soundly verifies signatures." A target deployed with `EXAMPLE_SIG_CHECKER_ADDRESS` pointed at a permissive/mock checker that happens to expose `registryCoordinator()` would still deploy, and such a target accepts unsigned diffs = full compromise. That's fine for a **dev/test harness deploying example contracts to a fork/testnet**, which is what this is — but it's worth a line in `scripts/README.md` stating plainly that this tool is not for deploying value-bearing production targets, so no one reaches for it that way later. If it ever needs to, the checker check would want to be authenticity-based, not just shape-based.

### Everything else lines up
`onchainLife`-requires-`prestate-net` is documented and now actually proven (branch rebased onto merged #348; the verification comment shows a full quorum round settling with anvil memory flat at ~30 MiB, versus the struct-log path's OOM — which is also what proves the net form ran rather than the canonical fallback). `fetch_examples.sh` pins the examples repo to a SHA (reproducible). The stale-`#163`-description correction in the body is a good catch and matches what actually merged.

Nice piece of tooling — it turns "add a new example" into a manifest edit, and the checker/submit sharp edges are the ones that mattered.

bagelface:
Thanks — took the scope caveat, in `a824474`.

You're right that the check is shape-based, and re-reading it my own comment was overclaiming: `validate_sig_checker`'s doc headline said it "rejects a signature checker that cannot actually verify signatures," which is not what it establishes. It establishes *has code and answers `registryCoordinator()`* — enough to catch the retriever and the examples repo's `MockBLSSignatureChecker` default, but a permissive checker exposing that getter passes, and a target wired to one accepts unsigned diffs.

Two changes:

- **`scripts/README.md`** — a blockquote at the top of the section, not buried at the bottom: dev/test only, do not deploy a value-bearing target, with the reason stated (shape vs authenticity, and that the failure mode is accepting unsigned diffs). Also notes what a production path would actually need — the checker verified against the registry coordinator the AVS registered against.
- **`scripts/deploy_example.rs`** — retitled the function doc to "rejects a signature checker that is *demonstrably* not one" and spelled out the shape/authenticity distinction at the definition, so someone reading the function doesn't trust it further than it goes. Comment-only; no code touched.

I put it in both places deliberately: the README stops someone reaching for the tool, and the function doc stops someone extending it on the assumption the check is stronger than it is.

Nothing else from the review needed action. `fmt`/`clippy -D warnings` clean and the 33 `deploy_example` tests still pass locally; CI re-running.

PR #348, opened 2026-08-07
feat: adopt the prestate-net state encoding from gas-analyzer

bagelface:
## Why

`onchainLife` — the Conway's Life example from #347 — deploys and passes the router's routability gate, but its task never renders. One generation is ~16.9M gas, and the struct-log extractor's `debug_traceCall` produced a trace large enough to **OOM-kill the anvil container** (`exit 137`, `oomkilled=true`):

```
ERROR gas_killer_router::sequencer: failed to enrich task, dropping request
  error=Gas analysis failed: debug_trace_call failed: error sending request
```

gas-killer/gas-analyzer#163 + #165 have merged and address exactly this: a net-diff extraction read from `prestateTracer` + `callTracer` costing `O(changed slots)` instead of `O(execution steps)`.

## Heads-up: #163's description is stale

Its body says "no public API change… the service picks this up on a dependency bump with zero code changes." That was true early on the branch and was **deliberately reversed before merge** by `60ab3f44` ("make the prestate net diff an explicit StateEncoding rather than an implicit fast path"); the body was never updated. The merged design:

- `StateEncoding::PrestateNet` is a **third, opt-in variant**, not an automatic fast path.
- Under `Legacy`/`Canonical` the prestate branch is never entered — same single `debug_traceCall`, same bytes. **The bump alone is behaviour-neutral.**
- A tracer failure under `PrestateNet` is a **hard error, never a quiet downgrade**. An operator whose node lacks the tracers would otherwise sign a digest no one else can match, turning a node-config difference into a consensus split.
- The service's `parse_state_encoding` **panics** on unrecognized values, so it had to learn the variant before the option was reachable at all.

(I repeated the stale claim in #347's description; that needs correcting too.)

## What this does

Adds the capability, leaves the default alone.

| | Eligible call | Ineligible call | Node lacks tracers |
|---|---|---|---|
| `legacy` (default) | struct-log, depth-1 SSTOREs | same | n/a |
| `canonical` | struct-log, checkpointed | same | n/a |
| `prestate-net` **(new)** | net form: one slot-sorted store per changed slot + logs | falls back to **canonical**, not legacy | **hard error** |

- **`Cargo.lock`** — `cargo update -p gas-analyzer-evmsketch`, `24051bce → 964eb229`. Exactly the four sibling-package `source =` lines and nothing else; `Cargo.toml` untouched, matching the established playbook (`d6782385`, `bc387a0` — a `rev =` pin has never been used here). 22 commits / 12 files ride along: #163 and #165 (library) plus #179/#180 (CI lint matrix, toolchain pin). No new dependencies, no declared MSRV, unchanged signatures and error type.
- **`common/src/config.rs`** — the `prestate-net` arm, updated docs, and tests. One accepted spelling per encoding: alternates would let two operators believe they are configured identically while signing different bytes, so `prestate_net` is rejected rather than resolved.
- **`example.env`** — a third bullet covering the net form, the canonical fallback, and the tracer requirement.
- **Helm** — `global.stateEncoding` (default `"legacy"`) threaded into the node and router deployments. It was previously unset chart-side, so clusters silently got legacy with no way to opt in. Render-verified for both default and override.
- **e2e matrix** — a fourth leg. `ArraySummation.sum` makes no external call at target depth, so it exercises the net form itself rather than its canonical fallback.

Deliberately **not** changed: `Cargo.toml`, `common/src/validator.rs` (its single call site passes `self.state_encoding` straight through), `docker-compose.yml` and `scripts/run_e2e_test.sh` (both already thread the value).

## Consensus note

`StateEncoding` selects the *signed representation*, not just how it is obtained — the encoded updates are hashed into the task digest. It must be uniform fleet-wide and rolled out atomically; a rolling update that transiently mixes encodings fails quorum until it converges. `legacy → prestate-net` changes digests for essentially every call, which is exactly why the default stays `legacy` here.

Enabling it anywhere beyond a local fork also requires every node's **and** the router's RPC to support `debug_traceCall` with `prestateTracer` (diffMode) and `callTracer` (withLog). Anvil supports both; hosted providers often gate `prestateTracer` to paid tiers.

## Testing

`cargo fmt`, `clippy --all-targets --all-features -D warnings`, `cargo check`, `cargo test` all green (253 tests). Helm templates rendered for both the default and the override.

**Draft** until the new `prestate-net` e2e leg goes green — the encoding's live behaviour is unproven until then. Local end-to-end verification was blocked by a full disk / wedged Docker on my machine, so CI is the first real exercise of the path.

## Follow-ups

- Correct #347's description and, once this lands, rebase it and rewrite its "Known limitation" section: `onchainLife` works under `STATE_ENCODING=prestate-net`, not automatically.
- `router/Cargo.toml` and `scripts/Cargo.toml` both declare `gas-analyzer` but never use it; `common` is the only real consumer. Unrelated cleanup.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -30,6 +30,10 @@ jobs:
         #   - the aggregate-Schnorr path under the CANONICAL encoding settling a task that
         #     makes a re-entrant external call mid-transition — a green run proves
         #     re-entrancy settles safely end-to-end.
+        #   - the BLS path under the PRESTATE-NET encoding, which signs the target's net
+        #     storage diff read from prestateTracer/callTracer rather than a struct-log
+        #     trace. `ArraySummation.sum` makes no external call at target depth, so this
+        #     leg exercises the net form itself rather than its canonical fallback.
         include:
           - signature_scheme: bls
             state_encoding: legacy
@@ -40,6 +44,9 @@ jobs:
           - signature_scheme: schnorr
             state_encoding: canonical
             e2e_example: reentrant
+          - signature_scheme: bls
+            state_encoding: prestate-net
+            e2e_example: array-summation
 
     steps:
       - name: Checkout code
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4015,7 +4015,7 @@ checksum = "42012b0f064e01aa58b545fe3727f90f7dd4020f4a3ea735b50344965f5a57e9"
 [[package]]
 name = "gas-analyzer-core"
 version = "0.1.0"
-source = "git+https://github.com/gas-killer/gas-analyzer#24051bce1251b890f0f2f4f0141026bfb52aca59"
+source = "git+https://github.com/gas-killer/gas-analyzer#964eb229cabb7c3d933b58ab2bf4f82c9be027cd"
 dependencies = [
  "alloy-primitives",
  "alloy-rpc-types",
@@ -4030,7 +4030,7 @@ dependencies = [
 [[package]]
 name = "gas-analyzer-estimator"
 version = "0.1.0"
-source = "git+https://github.com/gas-killer/gas-analyzer#24051bce1251b890f0f2f4f0141026bfb52aca59"
+source = "git+https://github.com/gas-killer/gas-analyzer#964eb229cabb7c3d933b58ab2bf4f82c9be027cd"
 dependencies = [
  "alloy-dyn-abi",
  "alloy-primitives",
@@ -4046,7 +4046,7 @@ dependencies = [
 [[package]]
 name = "gas-analyzer-evmsketch"
 version = "0.1.0"
-source = "git+https://github.com/gas-killer/gas-analyzer#24051bce1251b890f0f2f4f0141026bfb52aca59"
+source = "git+https://github.com/gas-killer/gas-analyzer#964eb229cabb7c3d933b58ab2bf4f82c9be027cd"
 dependencies = [
  "alloy",
  "alloy-eips",
@@ -4071,7 +4071,7 @@ dependencies = [
 [[package]]
 name = "gas-analyzer-rpc"
 version = "0.1.0"
-source = "git+https://github.com/gas-killer/gas-analyzer#24051bce1251b890f0f2f4f0141026bfb52aca59"
+source = "git+https://github.com/gas-killer/gas-analyzer#964eb229cabb7c3d933b58ab2bf4f82c9be027cd"
 dependencies = [
  "alloy",
  "alloy-eips",
--- a/common/src/config.rs
+++ b/common/src/config.rs
@@ -397,15 +397,27 @@ fn parse_signature_scheme(value: Option<&str>) -> SignatureScheme {
 }
 
 /// Reads the storage-update encoding from `STATE_ENCODING` (case-insensitive
-/// `legacy` | `canonical`), defaulting to [`StateEncoding::Legacy`].
+/// `legacy` | `canonical` | `prestate-net`), defaulting to [`StateEncoding::Legacy`].
 ///
 /// `canonical` emits canonical state slices before every external call and a
 /// final re-assertion slice, so a re-entrant call into a Gas Killer target
 /// observes the storage native execution would have had (see
-/// `gas_analyzer::compute_state_updates_canonical`). **Node and router must run
-/// the same value** — the encoding changes `storage_updates`, hence the task
-/// digest — so this is read from one env var threaded to both binaries.
-/// Panics on an unrecognized value rather than silently diverging digests.
+/// `gas_analyzer::compute_state_updates_canonical`).
+///
+/// `prestate-net` signs the target's *net* storage diff — one slot-sorted store per
+/// changed slot plus its logs — read from the `prestateTracer`/`callTracer` pair
+/// instead of a struct-log trace. It costs `O(changed slots)` rather than
+/// `O(execution steps)`, which is what makes heavy-compute tracked functions
+/// extractable at all: their struct-log trace is large enough to time out or exhaust
+/// the node. Calls the net form cannot represent (an external `CALL`, `CREATE`, or
+/// `SELFDESTRUCT` at target depth) fall back to the *canonical* encoder, and a node
+/// whose RPC lacks either tracer fails the task rather than quietly downgrading —
+/// see [`gas_analyzer::StateEncoding::PrestateNet`].
+///
+/// **Node and router must run the same value** — the encoding changes
+/// `storage_updates`, hence the task digest — so this is read from one env var
+/// threaded to both binaries. Panics on an unrecognized value rather than silently
+/// diverging digests.
 pub fn state_encoding() -> gas_analyzer::StateEncoding {
     parse_state_encoding(env::var("STATE_ENCODING").ok().as_deref())
 }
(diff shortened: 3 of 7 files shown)


bagelface:
### Measured: why this encoding exists

Local Docker was too disk-starved to run the full AVS round, but the underlying claim is testable directly at the RPC layer. Deployed `OnchainLife` (glider seed) to a stock anvil — foundry 1.5.1, the same build the compose stack runs — and issued the exact tracer calls each encoding makes against `step(uint32) 1`.

The call itself is heavy: `callTracer` reports **gasUsed 16,552,726**.

**`prestateTracer` (diffMode) + `callTracer` (withLog) — what `PrestateNet` issues:**

```
prestateTracer : 0s, 863 bytes, 3 changed slots on the consumer
callTracer     : 0s, 578 bytes, 1 log, no target-depth frames, no revert
```

Eligibility mirrors `classify_prestate_eligibility`: root frame targets the consumer, no `error`/`revertReason`, no regular `CALL`/`CREATE`/`CREATE2`/`SELFDESTRUCT` at target depth, and the only other account touched (the sender) changes no storage — so **this call takes the net form**, not the canonical fallback. Three slots matches the example repo's "3-op diff" note.

**Struct-log tracer — what `Legacy`/`Canonical` issue, same call:**

```
t=0s   anvil RSS      8 MB
t=5s   anvil RSS   1524 MB
t=10s  anvil RSS   1986 MB
t=20s  anvil RSS   2231 MB   <- still climbing, still no response; killed
```

Never returned. At that growth rate it reaches the compose container's 3.82 GiB limit in roughly 35s, which is precisely the `exit 137 / oomkilled=true` failure that motivated this PR.

So: ~1.4 KB and instant, versus unbounded growth and no result. That is the `O(changed slots)` vs `O(execution steps)` difference, on real hardware.

Two things this also settles:
- **anvil supports both tracers.** Worth confirming because `PrestateNet` hard-fails rather than downgrading when they are missing.
- **The 1.96 Docker builder is fine.** gas-analyzer's new `rust-toolchain.toml` pins 1.97 for its own CI, but declares no `rust-version`, so it is not an enforced MSRV — the green `build (node, …)` and `build (router, …)` legs confirm it.

Still pending: the `bls / prestate-net / array-summation` e2e leg, which is what actually exercises the encoding through a full quorum round and on-chain settlement.

bagelface:
### All 10 checks green, and the new leg exercises the net form

`e2e-test (bls, prestate-net, array-summation)` passed — a full quorum round settling on-chain under the new encoding. The three pre-existing e2e legs and both Helm legs also passed, which is the behaviour-neutrality claim confirmed: `legacy` and `canonical` are unchanged on the bumped dependency. The Helm legs deploy the chart to a kind cluster, so they exercise the `global.stateEncoding` addition for real rather than just `helm template`.

**One thing worth checking rather than assuming.** A passing leg does not by itself prove the *net form* ran — had `ArraySummation.sum` been ineligible, the leg would have silently exercised only the canonical fallback, which the existing canonical leg already covers. The `extraction` field is a tracing span field and is not emitted at `RUST_LOG=info`, so the CI log cannot answer this.

Confirmed directly instead. Built `ArraySummation` from the pinned SDK, deployed it with the e2e's own constructor arguments (`100 / 1000 / 42`), and ran `callTracer` on the call `send_request` actually builds, `sum([0,1,2])`:

```
root type          : CALL  (targets the consumer)
target-depth frames: none
error/revertReason : None / None
logs               : 1

ELIGIBLE for net form: True
```

No regular `CALL`/`CREATE`/`CREATE2`/`SELFDESTRUCT` at target depth, root frame targets the consumer, no revert — so `classify_prestate_eligibility` returns `Eligible` and this leg signs the net form. Matches the source: `sum` is pure arithmetic over storage plus `emit SumCalculated`.

Ready for review.

BreadrichEngels (approved):
## Review: adopt the prestate-net state encoding

**Approve.** This wires the already-reviewed `StateEncoding::PrestateNet` (gas-analyzer #163/#165) into the service as a **third opt-in variant with the default unchanged**, and it carries every piece of the consensus discipline this knob needs. Merging is behaviour-neutral for production, and the one live proof — a full quorum round settling on-chain under the new encoding — is green.

### The two things that make this safe to merge, both verified
1. **The dependency bump is behaviour-neutral for the default.** `Cargo.lock` moves the four sibling `gas-analyzer` packages `24051bce → 964eb229` and *only* those four `source =` lines — no transitive dep churn (consistent with #163/#165/#179/#180 adding none). Per the merged design, `Legacy`/`Canonical` never enter the prestate branch, so their bytes are unchanged — and the green `bls/legacy`, `schnorr/legacy`, `schnorr/canonical`, and both Helm legs *prove* it: a changed representation on the bumped dep would have forked those digests and failed the legs. So the bump alone can't move any existing deployment's digest.
2. **The new leg genuinely exercises the net form, not the fallback.** This was the exact thing I'd have flagged — a passing `prestate-net` leg could have silently run only the canonical fallback (already covered by the canonical leg) if `sum` were ineligible. You pre-empted it: confirming directly that `sum([0,1,2])` classifies `Eligible` (root targets the consumer, no target-depth frames, no revert) means the leg signs the net form and settles it through a real round. That's the right verification.

### Config discipline is exactly right
`parse_state_encoding` gains one canonical spelling (`prestate-net`), panics on a near-miss (`state_encoding_rejects_a_near_miss_spelling` pins `prestate_net` → panic), and stays fail-closed on anything unrecognized. That's the property that matters for a digest-determining knob: an operator can't mistype into a *running-but-divergent* state — they get a hard failure instead. Threading `global.stateEncoding` (default `"legacy"`) into **both** the node and router deployments from the same global value guarantees the two binaries can't drift apart, which is the mismatch that would otherwise fail quorum. And the tracer-failure-is-a-hard-error / no-quiet-downgrade behaviour (inherited from #165) is correctly surfaced in the docs.

### Operational gate before anyone flips it (documented; reiterating as the checklist)
The PR's consensus note is spot-on and I have nothing to add to it — just the pre-enablement checklist for whenever `prestate-net` is turned on beyond a local fork:
- **Digest migration.** `legacy → prestate-net` changes essentially every call's digest, so it's an atomic fleet-wide flip (node + router together), never a rolling update — a transient mix fails quorum until convergence.
- **Every node's *and* the router's RPC must support `prestateTracer` (diffMode) + `callTracer` (withLog)**, or that node hard-fails the task. Your measurement confirms anvil does; hosted providers often gate `prestateTracer` to paid tiers, so this needs confirming per-operator.
- **Cross-node tracer-output equivalence** remains the one thing I'd still nail down before a heterogeneous-RPC (non-anvil) deployment: PrestateNet's determinism rests not just on the classifier but on the `callTracer` log `position`/`index` fields agreeing across client implementations (the second byte-identity surface from gas-analyzer#178). Anvil 1.5.1↔1.7.1 agreeing (from #165) is reassuring but isn't the geth/reth/hosted matrix.

### Minor
- The consensus-critical encoder is pinned only in `Cargo.lock` (git-dep on the default branch, no `rev =` in `Cargo.toml`) — matching the repo's established playbook, so fine here, but worth being aware that an unintended `cargo update` could drift it; the committed lock is the sole source of truth for "which encoder every binary builds." Not this PR's to change.

The measured RPC-layer evidence (prestate ~1.4 KB instant vs struct-log unbounded RSS → the exact `exit 137` OOM that motivated this) is a great artifact to have on the PR. Clean adoption — capability added, default untouched, and the discipline that keeps the digest deterministic is all present and tested.

bagelface:
Thanks — acting on the third checklist bullet, since re-reading my own docs they were wrong about it by omission.

**Cross-node tracer-output equivalence (now documented, `4623a29`).** My `example.env` and Helm comments said every node's and the router's RPC must *support* `prestateTracer`/`callTracer`. That's necessary but not sufficient, and I'd written it as if it were the whole requirement. Support says nothing about output agreement: per gas-analyzer#178, `ordered_target_depth_logs` takes ordering from the global log `index` when every log carries one and from `position` otherwise, and `index` isn't universally populated — so two operators on different clients can both satisfy "supports the tracers" and still order logs differently. Same end state, different bytes, forked digest. Both call sites now say that explicitly and scope the claim: `prestate-net` is proven on a single-client fleet (the all-anvil local stack), not a heterogeneous geth/reth/hosted one, until #178 picks one source and fails closed.

I also documented the `extraction = prestate_net | struct_log` span field you'd want for detecting a split fleet — with the caveat that it's a span field, not an event, so it won't show at plain `RUST_LOG=info`. That's not academic: it's exactly why I couldn't confirm from the CI log which path the new leg took, and had to establish `sum([0,1,2])`'s eligibility directly instead.

Deliberately **not** changed:

- **The `Cargo.lock`-only pin.** Agreed on the hazard and agreed it isn't this PR's to fix. Worth noting the bump here was `cargo update -p gas-analyzer-evmsketch` rather than a bare `cargo update`, which is why the diff is exactly the four sibling `source =` lines — the last analyzer bump (`d6782385`) used the bare form and swept `sp1-contract-call` along with it. If you want the encoder pinned harder than convention, that's a standalone change.
- **Nothing about the digest-migration or RPC-support bullets**, which were already covered.

No code changed, so the green run above still stands for the encoder itself; CI is re-running on the docs commit.

